PR #70, opened 2023-08-05
⚡️ Optimized `bs.math:sqrt`

Leirof:


Leirof:
This new sqrt doesn't pass the unit tests
image

Leirof:
I contacted the creator of the function to resolve this: https://discord.com/channels/154777837382008833/1121746379103027232/1137677866201534597
(Link to a message on Minecraft Commands discord server: https://discord.gg/QAFXFtZ)

aksiome:
In the info block Gunivers should be Gunivers Community. Also not sure about that but i think the author field should be a comma separated list. Some files does not end with a new line, not sure if we should enforce that but github on my phone is complaining about that.

aksiome:
Also there is an empty space in the function tag last line

aksiome (approved):

PR #87, opened 2023-08-14
Added bs.math:combine

Ethanout:


Leirof (approved):
Tested and approved ✅
Accepting PR as it is.

3 things remain to be done to fully complete this new feature:
- Create a documentation for this feature
- Add it to the changelog
- Replace commented debug tellraw by log functions

I will take care of all of that in a new PR because it's the boring part, I don't want to bother you with that for now. It's already hard enough for you to deal with git 🤣

PR #90, opened 2023-08-16
Update biome module to follow new conventions

aksiome:
I removed things that were unused or now outdated. `_.mcfunction` was moved to `__help__.mcfunction`, can't remember if it was what we said. I also added the help function tag and everything can be setup from the __load__ this way the module can be completly standalone. In the end I think this is something we should aim for even if we get some obj registration / constant score duplication.


aksiome:
I'll work on the documentation later, maybe we should wait for that before reviewing or merging 😉

aksiome:
Since the pull is quite big to review, here is the list of what changed:
- Since now we can execute if biome the get_id was now not really useful so it was remove
- The get temperature functions are basically the same they just use a return statement instead of a score
- can_rain and can_snow are now predicates and there is a new has_precipitation
- The is_raining was moved to the weather module instead of being in the biome one, i also added an is_thundering
- The biome displayer was removed from the module list since it was changed to only an example. I also removed the example since we now only need to `execute if biome <any> run title <>`
- metadata was put on the `__load__` instead of help
- the load define all scores needed, this way the module can completly be standalone which i think is something we should aim for

I hope i did not forget something. Hopefully this will help you review it faster

Leirof (approved):
LGTM

PR #91, opened 2023-08-17
Plot-system-V2

VForiel:
This PR add a better plot system in the sandbox dedicated datapack.

This new plot system is only asking to define:
- X limits
- resolution (number of points)
- plot origine (X, Y and Z)
- plot size (X, Y and Z)

Interpolations are made to assigne an X value to each entity and scale the Y range in function of the results. Create a plot can then be performed in few simples commandes (see bs.plot:sqrt)


VForiel:
You can close it

Issue #92, opened 2023-08-17
Add a sidebar module for dynamic scoreboard sidebar


aksiome:
This is now easily possible thanks to macros

PR #93, opened 2023-08-18
Custom sidebar module

aksiome:
Implements #92 and since I wanted to do some validation, instead of doing it inside the sidebar module I created a new validator module


aksiome:
I cannot find a good icon for this module. The one i chosed (📰) look a bit austere, maybe someone will get a better idea...

aksiome:
Also the validator module was removed for now, it was a too much for a single PR.

Leirof (changes requested):
Very cool!

The only problem I saw is that it miss some exemples in the doc (for delete, refresh and reset, even if it seems trivial, a copy/pastable exemple is always good) and information about how to update a sidebars (there is create and refresh, but not add/remove/update function, so how to add/remove lines from an existing sidebars?

aksiome:
It's currently not really possible to update/add/remove lines. Updating would be easy to make but i think it will add complexity to the format since we need to specify each line. And adding and removing is a bit more tedious. Imo it's not really worth it. In most case you're gonna call the create command inside an init function so users can just reset then create again which is a lot better if we consider what is really going on internally. That's also why i said in the doc than resetting is better than deleting a single sidebar.

Leirof (approved):

PR #97, opened 2023-08-23
⚡️`bs.block` Update bs.block

theogiraudet:
Update to remove explicit tree and use macro/storage instead.

- [x] block type from block 
- [x] blockstate from block
- [x] block nbt from block
- [x] all from block
- [x] to setblock
- [x]  to setblock with mode
- [x] to fill
- [x] to fill with mode
- [x] to fill with replace
- [x] to if block 
- [x] to unless block
- [x] to if blocks
- [x] to unless blocks
- [x] to item
- [x] from item
- [x] to falling block
- [x] to display block entity
- [x] to falling block with NBT
- [x] to display block entity with NBT
- [x] to block display with hitbox
- [x] to block display with hitbox with NBT
- [x] set NBT
- [x] set type
- [x] set state
- [x] blockstate merge 
- [x] blockstate cherry pick 
- [x] release the storage file


theogiraudet:
#resolves #9

RedCoal27:
Will you give the storage file instead of big function that Generate the storage for people who what the datapack for a map? 
So instead of having to manipulate big functions you can just download directly the storage

theogiraudet:
@RedCoal27 The Python generator is now able to generate the `command_storage_bs.dat` with the block state and block type lists.

aksiome:
New format is finalized. We still need the new generator on which @Leirof is working and the documentation. Here is a list of the features to review:


## Features

- `fill_block`*: a simple fill command
- `fill_type`*: try to swap the type without changing the state if state cannot be conserved the default one will be used instead
- `get_block`: will retrieve block information
- `get_type`: will only retrieve the type (available states are returned which mean transformation functions can be used, however no state is selected)
- `item_to_block`: will retrieve the type from an item
- `keep_properties`**: filter to keep only the desired properties
- `match`: works like an if block, to use with the if/unless function
- `merge_properties`**: will try to merge state properties of the current location onto the output. This operation is very permissive, if a value can work even if it makes no sense it will be merged (ex: age for different plants)
- `remove_properties`**: the opposite of keep_properties
- `replace_type`**: replace the ouput type. This function will keep the state when the state group is the same and will use the default state if not. (this function is internally used by set_type and fill_type)
- `set_block`: just set a block
- `set_type`: like for fill_type try to only set the type if not possible use type + default state
- `shift_properties`: shift any properties by any amount which allow cycling through properties
- `spawn_block_display`: spawn a block display with the given block
- `spawn_falling_block`: spawn a falling block with the given block
- `spawn_solid_block_display`: spawn a block display that has an hitbox.

\* These features are processes. They can be be configured to run on mulitple ticks in order to improve performances. They also support a masks syntax which is a lot more powerful than the default filter that minecraft has when using the replace mode.

** These functions are "transformation" functions. These functions are used to modify the output, since the user should never attempt to do it by himself. An other particularity is that they are the only ones that use macro params for simplicity (ex: `function #bs.block:remove_properties {properties:[{name:"facing"},{name:"shape"}]}`)


## Todo
- [ ] Update documentation
- [ ] Add comments & check file headers
- [ ] Finish the generator (with the ability to export the storage file?)*

\* With the new format this is something to discuss since it only stores all blocks and not all blocks with all states (the size is greatly reduced). Also the function avoid a lot of redundancy by using multiple data commands instead of a single one, so we should first check that the storage file is lighter than the function.

Leirof:

PR #98, opened 2023-08-23
Major issue

Leirof:


theogiraudet (changes requested):
Please, use the new log module instead.

Leirof:
Hmmm, what about no?

Leirof:
Not prioritary, neither my job

Leirof:
Current unit tests will be rewritten with the unit test module ( #106 )
This PR is then out of date even before being merged, so I close it ^^

PR #101, opened 2023-08-25
🚸 Added @s support in "message" macro

Leirof:
This pull request add the support of @s in the message.

Previously, logs behaved like this:
```
scoreboard players set @e[tag=yolo] bs.score 1
execute as @e[tag=yolo] run function bs.log:info { \
  path: 'osef', \
  feature: 'osef', \
  message: '{"text":"Score: "}, {"score":{"name":"@s","objective":"bs.score"}}' \
}
# [...] Score:
```
Because the tellraw was executed on the players that recieve the message, so the @s selector in the message doesn't take the right entity.

Now, it will work as expected
```
scoreboard players set @e[tag=yolo] bs.score 1
execute as @e[tag=yolo] run function bs.log:info { \
  path: 'osef', \
  feature: 'osef', \
  message: '{"text":"Score: "}, {"score":{"name":"@e[tag=yolo,limit=1]","objective":"bs.score"}}' \
}
# [...] Score: 1
```

### DIFF
--- a/datapacks/Bookshelf/data/bs.log/functions/debug.mcfunction
+++ b/datapacks/Bookshelf/data/bs.log/functions/debug.mcfunction
@@ -19,4 +19,8 @@ $execute as @a unless entity @s[ \
     tag=!bs.log._.debug, \
     tag=!bs.log._._, \
     tag=!bs.log.$(feature)._\
-] run tellraw @s ["",{"text":"BS","color":"aqua"},{"text":" [DEBUG]","color":"#CCCCCC"},{"text":" $(path)","color":"dark_aqua"},{"text":" > ","color":"gray"}, $(message)]
+] run tag @s add bs.log.catch_log
+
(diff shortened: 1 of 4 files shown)


datapacks/Bookshelf/data/bs.log/functions/debug.mcfunction
theogiraudet, line 1: The tag used in the tellraw doesn't match the given/removed tag

datapacks/Bookshelf/data/bs.log/functions/error.mcfunction
theogiraudet, line 1: Should use @a with the new introduced tag

datapacks/Bookshelf/data/bs.log/functions/info.mcfunction
theogiraudet, line 1: Should use @a with the new introduced tag

datapacks/Bookshelf/data/bs.log/functions/warn.mcfunction
theogiraudet, line 1: Should use @a with the new introduced tag

theogiraudet (changes requested):

aksiome (approved):

PR #103, opened 2023-08-25
Bitwise module update

aksiome:


Leirof (approved):
LGTM

I didn't test all the functions but you mostly change the interfaces so it should be ok. The thing is that when you move files, git consider you deleted the original one and created a new one, so everything seems changed. It exist a command (I don't remember the name) to properly move files in a git repo. Maybe there is a VSC extension or option to use this command instead of the one provided by the OS

PR #104, opened 2023-08-27
🎨 Update color module

Leirof:


Leirof:
Aksiome put the mess, it was originally his PR. I approve and merge it (not a yolo merge then).

PR #108, opened 2023-08-28
🎨 Fix consistency issues

aksiome:
- Missing file header in bs.color
- Missing shared objectives in bs.bitwise


RedCoal27:
Will you give the storage file instead of big function that Generate the storage for people who what the datapack for a map? 
So instead of having to manipulate big functions you can just download directly the storage

Leirof (approved):
LGTM

PR #109, opened 2023-08-29
⚡️ Update bs.id

### LINKED ISSUE #8: `bs.id`: Fix `update_cuid`

aksiome:
This PR breaks other modules but since i need this PR to update other modules and i don't want to create a huge PR update, I'll open an issue with the list of modules that it breaks when merged. This way we can safely keep track of them:

- bs.item
- bs.link
- bs.view
- some unit tests

This PR introduces predicates instead of the check functions, changes the `get_<id>` functions to `give_<id>` since it not only get but generate then set a score, and also fixes #8 !


Leirof (approved):
LGTM

PR #112, opened 2023-08-30
🐛 Just simple working systime

aksiome:
⌚ Hop hop hop au boulot @Leirof 

PS: wtf is sidebar load doing here


aksiome:
We can't really change the output, it's currently based on your system so if your server is in GMT you'll get GMT. And we can't really detect the timezone... So it's up to the user to know the current configuration of his system!
The doc is better for sure :thumbsup:

Leirof:
Oh ok, weird, our server is then on UTC-5 🤔

Leirof (approved):

PR #114, opened 2023-09-05
Update math module

aksiome:


aksiome:
I tryied to be coherent when naming things but maybe everything is not the best don't hesitate to tell me if i should rename a fake score or storage entry. For example i was'nt sure between .x and .value for functions?

docs/modules/math.md
```diff
@@ -242,620 +247,586 @@ Examples
...
 
-:  (execution) `as <entities>`
-   : The entities you want to perform the operation on
+:  (score) `$math.pow.base bs.in`
+   : The base.
 
-   (score) `@s bs.in.0`
-   : The number to be logarithmized, shifted by 3 digits (1,2345 -> 1234) for better precision in integer scores
+   (score) `$math.pow.exp bs.in`
+   : The exponent.
```
Leirof, line 524: Missing scale input

Leirof (approved):

PR #118, opened 2023-09-09
📝 🎨 Update the schedule module

theogiraudet:
Update for the schedule module with three features:
- Schedule a command
- Cancel a scheduled command
- Cancel all scheduled commands


Leirof:
Merged in #119

PR #119, opened 2023-09-10
Update schedule module

aksiome:
Based on @theogiraudet  PR #118 !
Based on some not very accurate tests (i just used a lot of different random scheduled operation in a loop and compared debug charts) it should be at least 2 times faster, maybe even 3 😁


datapacks/Bookshelf/data/bs.schedule/functions/cancel_all.mcfunction
```diff
@@ -0,0 +1,20 @@
+# INFO ------------------------------------------------------------------------
+# Copyright © 2023 Gunivers Community.
+
+# Authors: theogiraudet, Aksiome
+# Contributors:
+
+# Version: 2.0
```
theogiraudet, line 7: The remove_all feature didn't exist before the PR so this is the ﬁrst version if this feature.

theogiraudet (changes requested):
I don't see documentation update

Leirof (approved):

Leirof (approved):

PR #120, opened 2023-09-10
:recycle: `bs.link` Update link module

aksiome:
Almost done, just missing the documentation.


aksiome:
This PR also fixes bugs introduced inside the position module and add 2 new predicates to the id module: has_cuid and has_suid !

Leirof (changes requested):
Mirror point seems to not work properly

I put an AS on 3 100 0 and I place myself in -3 100 0

I link it to me and I run
`execute as @e[type=armor_stand,scores={bs.link.to=1..}] positioned 0 100 0 run function #bs.link:mirror_point_ata`

The AS goes very far (~100 blocs), but well keep a position on the opposite side of the point

Leirof (changes requested):
There is still resilience issues when scores or entity are cleared

aksiome:
I'm curious about scores, what is the problem with them? For the entity problem i dont wanna fix it right here because it concerns other module. I was planning to fix all modules that rely on global entities in an other PR later on!

PR #125, opened 2023-09-17
:truck: Change datapacks structure

aksiome:


Leirof:
Also, wrong file arborescence here
image

aksiome:
The config datapack is not of my doing x). Anyhow the move module hasn't been updated yet and if we keep a config like datapack, I think it's easier to put it inside the examples. Also I completly forgot about the example datapack I'll change that right away.

Leirof (changes requested):
- bs.move in Bookshelf Example should actually be in another datapack imo such as "Bookshelf Config", "Bookshelf Modifiers" or "Bookshelf Addons"
- bs.map_edit should be in the main datapack, idk if it's already the case (or if it simply work), but the user can use it to edit the map during a game, it is not only designed for the dev phase (or it shouldn't be)
- bs.unit_tests should be moved in Bookshelf Dev as it have nothing to do with the world, it is made to check if the lib work as expected (and in the future, it will also help the dev see if he's own systems work properly)

aksiome:
These are kind of valid points. I'd really like @theogiraudet 's take on this too.
- The thing with bs.move is that the module is gonna be reworked and i'd really like to avoid having to rely on an additional pack since we can now use macros. I just put it there for now to keep a trace of it.
- For map_edit I think that we should split the module in half meaning the interface should be inside Dev and commands that are actually useful for generation and updating the map into the main datapack. Since this module should also be updated i put it into dev but I don't rely mind putting it back to where it was for now.
- With unit_tests the goal is to have a new module in dev that is used to create unit_tests but I think the tests that we write to check if the library works should not be inside the same folder. That's why i put it inside the world folder but maybe we should consider an other folder named tests. The only downside of this is that some tests may need to be based on the world, that's an other reason why i put it into the World. Maybe Sandbox was a better name? And if we prefer Tests since plot is the only thing of World it could be moved to Tests?

### We could have this structure:
- Bookshelf Dev
- Bookshelf Examples (could also be named Demo?, not sure if it's better tho)
- Bookshelf (Tests or World or both)
- Bookshelf

theogiraudet:
> We could have this structure:
Bookshelf Dev
Bookshelf Examples (could also be named Demo?, not sure if it's better tho)
Bookshelf (Tests or World or both)
Bookshelf

I agree with this structure.

PR #127, opened 2023-09-18
fix: place the version switcher in the navbar-start

12rambau:
as mentioned in https://github.com/pydata/pydata-sphinx-theme/issues/1458, the header navbar goes on 2 row because the version-switcher is not an inline component anymore. I replaced it in the header-start to avoid it.


Leirof:
Oh ok, thx!

Leirof (approved):

PR #128, opened 2023-09-18
fix: rollback to simple search icon

12rambau:
The search-icon-field is to wide for your documentation making everything to go on 2 lines, let's see if rolling back to the regular button can be a nice workaround before we found a solution on our side.


Leirof:
Weird, it work well on local with the search bar. Why is there a difference with the RTD version?
image

But yes, let's try this

Leirof (approved):

Leirof:
Oh, I see, the dark theme button is not visible on the local version

PR #131, opened 2023-10-01
🐛 `bs.math` Fix exp function

aksiome:
- negative values that were outside the range should now return 0 instead of looping back to bigger ones


Leirof (approved):
LGTM

PR #152, opened 2024-01-20
⚡️ `bs.schedule` Improve module

### LINKED ISSUE #144: Improve schedule module
## Tasks
- [x] Remove selector param and instead keep the current context (both location and selector)
- [x] Use schedule command instead of a ticking function

aksiome:
closes #144


Leirof:

PR #154, opened 2024-01-21
🎨🐛 `bs.move` Improve module 

### LINKED ISSUE #143: Improve move module
## Tasks
- [x] Use $move.hit_face & $move.vel_remaining.[x,y,z] bs.out instead of obscure scores on bs.data for resolution (this is to avoid breaking changes and have a real interface for custom collision)
- [x] ~Improve recursion performances (see if it's possible to use the same method as in the block module)~ This will certainly not improve performances, current implementation already use a single function per recursion. Also using multiple axis functions (xyz, yz, and z) saves commands. We also avoid using additional macros, and everything should be cached (having more than 8 different ignored_blocks/solid_entities configuration seems very unlikely).

aksiome:
Closes #143


Leirof:

PR #156, opened 2024-01-29
⚡️ `bs.math` Improve accuracy and simplify things

aksiome:


aksiome:
I think this PR is good for now. I managed to optimize `sqrt` to make it almost as efficient as the old function. I changed the I/O because it wasn't ideal, and if we want to avoid breaking changes in the future, it's better to do it right now. I believe it will be crucial to review the I/O of the entire library before the release to ensure everything is consistent.

## Changelog

- I/O based on C / Python's documentation, causing many breaking changes (.angle => x, .value => x, combine.[m,n] => combine.[n,k], ...)
- Renamed the function `pow` to `ipow` (to be more consistent with `sqrt` and `isqrt`).
- Deleted `pow2` (which should have been renamed to `ipow2`) since the` ipow` function is already very performant.
- Added a `pow` function that supports float base and exponent (`sqrt` remains much more efficient than `pow(x,0.5)` 😜).
- Added an `exp2` function (2^x) that allows for better calculation of `exp` and avoids the lookup table. Also, the limit is now that of floats (e.g., 10^38), so for `exp2` bounds are [-127,127] instead of [-6,15] before with `exp`.
- The precision of almost all functions is much greater (and performance too, hopefully), except perhaps the `pow` function, which quickly loses precision as it relies on the fact that `log2` must be extremely precise: `pow(x,y) = exp2(y*log2(x))`.

PR #160, opened 2024-03-09
New bs.block generator

Leirof:


Leirof:
giphy

aksiome (approved):
Merge now since it works and will refine in other PR later

PR #183, opened 2024-05-03
Add capability to create new message log format

### LINKED ISSUE #170: Enable modification of the log message
If a data packer creates a new library over Boohshelf, `bs.log` may be used to log messages to the user of the new library.
In this situation, the data packer should be able to modify at least the prefix of the log message to replace it by the name of their own library.
In the following screenshot, the prefix is "BS" for "Bookshelf".

image

Proposition:
- Add a new storage `bs:config` where the data packer can overwrite some default configuration.
- Add `bs:config log.prefix.text`, `bs:config log.prefix.color`, `bs:config log.prefix.feature.color`, `bs:config log.separator.text`, `bs:config log.separator.color` and why not ways to configure the color of the severity.

theogiraudet:
This PR resolves #170.


aksiome:
Do we really have to set the log format on each load  module? Like can we not just use the default config for all of bookshelf. Also I'm not sure im a fan of having a default namespace, that means the namespace called "default" cannot really exists and we're exposed a little bit to unintended side effects. Maybe just hardcoding the default format inside each log is simplier? Last thing, personally i think we should consider this format instead:
```mcfunction
data modify storage bs:const log.messages set value [{namespaces:["anamespace","anotherone"],format:'["textcomponent here"]'}]
# we could use this syntax to add a new one
data modify storage bs:const log.messages[{namespaces:["mynamespace"]}].format set value '["textcomponent"]'
# or even add a namespace to a format
data modify storage bs:const log.messages[{namespaces:["mynamespace"]}].namespaces append value "otherone"
```
This way we can easily define multiple namespaces for a single format.

theogiraudet:
> Do we really have to set the log format on each load module?
It is what we discussed on Discord yeah. If we want to have a special log format for the whole Bookshelf, we need to redefine it for each module using the logger. It's also a good example of how to create a customized (complex) log format for interested people.

> Also I'm not sure im a fan of having a default namespace, that means the namespace called "default"
I can move the default formats to another path to enable the "default" namespace.

> Maybe just hardcoding the default format inside each log is simpler?
I think it is better to have a system like that:
- The default message formats may be override
- This simplifies future updates of the formats by having a single file to modify instead of four.

> data modify storage bs:const log.messages set value [{namespaces:["anamespace","anotherone"],format:'["textcomponent here"]'}]
Here you forgot to specify the level.
I assume you mean:
```
data modify storage bs:const log.messages set value [{namespaces:["anamespace","anotherone"],format: {info: '["textcomponent here"]'}}]
```
If yes, so why not.

aksiome:
> It is what we discussed on Discord yeah. If we want to have a special log format for the whole Bookshelf, we need to redefine it for each module using the logger. It's also a good example of how to create a customized (complex) log format for interested people.

But do we really need a special one? Having an example is nice tho. Also with the array version it's a lot easier to add each namespace to the bs format.

> I can move the default formats to another path to enable the "default" namespace.

Maybe if we use the array version we can have the first entry be the default one without even specifying a namespace, and/or having an other key like default:1b to target it (who knows some people may prepend to the array)

> Here you forgot to specify the level.
> I assume you mean:

Yes exactly did not think of the level but it work the same

theogiraudet:
> But do we really need a special one? Having an example is nice tho. Also with the array version it's a lot easier to add each namespace to the bs format.

I think so. It makes it easy to see which logs are from Bookshelf and which are not.

aksiome (approved):

PR #198, opened 2024-05-25
✨ feat: add new painter example

aksiome:


examples/paintbrush.md
```diff
@@ -0,0 +1,306 @@
...
+
+## 📦 Requirements
+
+Before we start, make sure you have:
+- Minecraft Java Edition 1.21.4 or later
+- Bookshelf `bs.block` and `bs.view` modules, see [Quickstart](https://docs.mcbookshelf.dev/en/latest/quickstart.html) for installation
+  - `@require bookshelf.module.block`
+  - `@require bookshelf.module.view`
+- Basic knowledge of datapack development, including functions, tags, and scoreboards
+
```
theogiraudet, line 25: Perhaps we should add a download to an empty datapack template to quick start, or a "- a ready-to-use empty datapack"?
Because we directly start by "create a function".

examples/paintbrush.md
```diff
@@ -0,0 +1,306 @@
...
+---
+
+### 4. Paint the Targeted Block
+
+Now, let’s move on to painting! We’ll use the `bs.block` module to apply the block type from the item held in the offhand (slot -106) to the targeted block. This process involves three steps:
+
+1. Load the targeted block into storage (`bs:out block`) as a "virtual block"
+2. Transform the virtual block using various operations (`mix_type` in this case)
+3. Generate the final result, which can be a block, item, particle, block display, etc. (here, we’ll replace the block with its transformed version).
```
theogiraudet, line 164: "Generate the final result from the virtual block, …"

examples/paintbrush.md
```diff
@@ -0,0 +1,306 @@
...
+
+---
+
+#### Set Up the Transformation
+
+The `mix_type` function uses a mapping registry to determine how blocks are transformed. Bookshelf provides two built-in registries:
+- `bs.colors`: This registry is used to transform blocks based on color properties. For example, if the held item is a colored block, it will attempt to apply the color transformation to the targeted block.
+- `bs.shapes`: This registry is used to transform blocks based on their shape. For instance, if the held block is an oak plank, and the targeted block is a birch stair, this registry would map the oak plank to a corresponding block shape like oak stairs. This is ideal for transforming between blocks that have a similar structural shape.
```
theogiraudet, line 172: "oak plank**s**", "bich stair**s**"

theogiraudet (changes requested):

examples/paintbrush.md
aksiome: Ideally a template would be best, could be what we talked about a few days ago (a template with beet and bookshelf preinstalled). But I think it's out of this scope and something for later, also it may be too complex for small experimentation, so i'll just use a new bullet item for now

theogiraudet (approved):

PR #218, opened 2024-06-15
Enhance speed of factorial function

SBtree-bit:
# Tasks to do before merging
- ✅ I accept to publish my contribution under [MPL v2](http://mozilla.org/MPL/2.0/) License
- ✅ My pull request is linked to an existing issue
- ✅ I have added my contribution to the [changelog](https://github.com/Gunivers/Bookshelf/blob/master/docs/CHANGELOG.md)


aksiome (approved):
I don't really know why but Leirof wanted to keep the old behavior and let the scoreboard overflow on bigger numbers so i added back the original code at the end of the function and updated the changelog for the 2.2.0 !

PR #228, opened 2024-06-20
🎲 Implement `bs.random` module!

SBtree-bit:
# Tasks to do before merging
  - [x] I have [tested](https://bookshelf.docs.gunivers.net/en/latest/contribute/debug.html#unit-tests) my contribution.


datapacks/Bookshelf/data/bs.random/function/noise/value_noise/utils/sin.mcfunction
aksiome, line 1: You should not use fake scores from other modules. It's kind of fine for the whole `bs.data` part, and even the `bs.out` part, but in my opinion, it should be discouraged (just use a bunch of other fake players on `#random.<value> bs.data`). For `bs.in`, this is blocking since we have a strict policy of preserving user inputs.

datapacks/Bookshelf/data/bs.random/function/noise/value_noise/value_noise.mcfunction
aksiome, line 1: I think in this case it's better to combine macros, also avoiding using data merge. Here's how I would write it. I guess this also apply for the white noise. (I just wrote it like that so i'm not sure the syntax is perfect, also there's a range in the doc that is not present here?):
```mcfunction
$data modify storage bs:ctx _ set value {tile_size:$(tile_size),height:$(height),width:$(width),spacing:$(height),postpone:0}
$data modify storage bs:ctx _ merge value $(with)

execute store result score #random.tile_size bs.data run data get storage bs:ctx _.tile_size
execute store result score #random.height bs.data run data get storage bs:ctx _.height
execute store result score #random.width bs.data run data get storage bs:ctx _.width
execute store result score #random.spacing bs.data run data get storage bs:ctx _.spacing
execute store result score #random.postpone bs.data run data get storage bs:ctx _.postpone

scoreboard players set #random.x bs.data 0
scoreboard players set #random.y bs.data 1

data modify storage bs:out random.noise set value []

function bs.random:noise/value_noise/xloop

# #result and #success are both contextual scores that we can use in bs.data.
# Basically, this means that if you are not calling any other feature and their use is controlled, you can use them.
# The benefit is to avoid spamming the scoreboard with too many entries.

execute store result score #result bs.data run random value 1..10000
scoreboard players operation #random.seed bs.data += #result bs.data
```
and instead of calling `exec` which personally i would rename `run`, (but that's not a convention thing, just a lot of callback functions are already named this way in the lib):
```mcfunction
function bs.random:noise/value_noise/run with storage bs:ctx _
```

```mcfunction
$function $(function)
data remove storage bs:out random.noise
data modify storage bs:out random.noise set value []
```

aksiome (changes requested):
Amazing work here! The logic and overall structure of the internals look really good to me, but there are a few convention-related things to fix:
- We should avoid creating custom objectives for fake scores to minimize the footprint of Bookshelf (`<value> bs.random` can be replaced by `#random.<value> bs.data`).
- Similarly we cannot create custom storage at all. Instead, use `bs:const random`, `bs:data random`, and `bs:ctx |y|x|z|_]` for fast contextual things.
- All features should be at the root of the module and match their input/output names. For example, with `#bs.random:distributions/binomial`, the output is `$random.binomial`, which does not match the name of the feature. I guess it's fine to move `#bs.random:distributions/binomial` to `#bs.random:binomial`. This way, we avoid long names and, actually, sorting by directory doesn't really change anything for the user since it behaves the same as an _ for autocompletion.

datapacks/Bookshelf/data/bs.random/function/__load__.mcfunction
```diff
@@ -0,0 +1,65 @@
...
+scoreboard players set #12 bs.const 12
+scoreboard players set #33 bs.const 33
+scoreboard players set #78 bs.const 78
+scoreboard players set #100 bs.const 100
+scoreboard players set #200 bs.const 200
+scoreboard players set #400 bs.const 400
+scoreboard players set #1000 bs.const 1000
+scoreboard players set #10000 bs.const 10000
+scoreboard players set #log(10) bs.const 23026
+scoreboard players set #exp(-0.9)*65536 bs.const 26645
```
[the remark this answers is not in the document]
theogiraudet: It doesn't matter to me.

datapacks/Bookshelf/data/bs.random/.metadata/module.json
```diff
@@ -0,0 +1,12 @@
...
+    "name": "bs.random",
+    "display_name": "Random",
+    "description": "Contains many random functions for different distributions",
+    "documentation": "https://bookshelf.docs.gunivers.net/en/latest/modules/random.html",
+    "authors": [
+        "SBtree"
+    ],
+    "contributors": [
+        "Aeldrion"
+    ]
```
theogiraudet, line 11: Since the author and contributor are already specified in at least one feature metadata, you don't have to define them here. Indeed, our scripts can extract this information from the feature tag metadata.
In practice, the authors/contributors part of the `module.json` is mainly for modules where we have features that are not functions. For instance, the `bs.tree` module is just a bunch of structures and we cannot define the authors of these structures inside them since they are not text files.

datapacks/Bookshelf/data/bs.random/tags/function/choose.json
```diff
@@ -0,0 +1,22 @@
+{
+    "__bookshelf__": {
+        "feature": true,
+        "documentation": "https://bookshelf.docs.gunivers.net/en/latest/modules/random.html#random-choice",
+        "authors": [
+            "SBtree"
+        ],
+        "contributors": [
+        ],
```
theogiraudet, line 9: If there are no contributors, you can remove this pair.

datapacks/Bookshelf/data/bs.random/tags/function/noise/value_noise.json
```diff
@@ -0,0 +1,22 @@
+{
+    "__bookshelf__": {
+        "feature": true,
+        "documentation": "https://bookshelf.docs.gunivers.net/en/latest/modules/random.html#noise",
+        "authors": [
+            "SBtree"
+        ],
+        "contributors": [
+        ],
```
theogiraudet, line 9: If there are no contributors, you can remove this pair.

docs/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Bookshelf is now based on **Minecraft 1.21**.
 
 ### 🎉 Addition
 - ([187](https://github.com/Gunivers/Bookshelf/issues/187)) Add `#bs.block:map_type` and `#bs.block:mix_type` functions.
+- ([1](https://github.com/Gunivers/Bookshelf/issues/1)) ([220](https://github.com/Gunivers/Bookshelf/issues/220)) Add `bs.random` module for random distributions and noise generation
```
theogiraudet, line 9: You just forgot a dot at the end of the sentence. :D

theogiraudet (changes requested):

aksiome:
For the documentation a lot of nbt have the wrong types. On top of that a lot of sentences are missing a dot at the end and ':' for examples.
Also we discussed a bit with Theo about interfaces and here is what we think should be best. If you have better ideas or you think something is bad don't hesitate to tell us. Naming things can be hard and we're always open for discussion.

## `#bs.random:binomial`

[inputs]
trials: [int]
probability: [double] (instead of chance? and use a double in range [0,1] instead of a scaled number, it can easily be stored then retrieve from the storage in a scaled way, but it's easier to understand for the user)

## `#bs.random:geometric`

[inputs]
probability: [double] (same as before double in range [0,1] instead of a scaled number)

## `#bs.random:poisson`

[inputs]
lambda: [double] (same as before, better not to use a scaled number)

## `#bs.random:choose`

[inputs]
bs:in random.choose.options (instead of bs:in random.choose.list)

[outputs]
bs:in random.choose (instead of bs:out random.choose.selection)

## `#bs.random:white_noise`

[inputs]
width: [int]
height: [int]
range: (remove it? it's unused in the code, or maybe i missed something)
with:
	callback: [str] (instead of function, also even if it's not really useful we think it should accept a command and not just a function)
	spacing: [int] (i think it should be the number of tiles to generate instead of rows) (should have a default value this way dont need to define it if a callback is given)
	postpone: [bool] (should have a default value (not sure which is best, maybe false?))

[outputs]
bs:out random.white_noise (instead of bs:out random.noise)
A two-dimensional array of 0s and 1s, with the size of width by height. (Not really clear it's a two dimensional array of values in range 0,1 right?)

## `#bs.random:value_noise`

[inputs]
width: [int]
height: [int]
scale: [int] (instead of tile_size. i like the simple description tho, it's easier to understand that talking about frequency)
with: Same as white noise

[outputs]
bs:out random.value_noise (instead of bs:out random.noise)

Example
Sadly the example cannot be run easily because it requires creating a function. I think this is something that could be added to the Examples datapack but this is something we can see later.

Note
A good thing with the with keyword, is that in the futur we'll be able to add things such as octaves to generate more complex terrain, without introducing breaking changes

aksiome (approved):

docs/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### 🎉 Addition
 - ([187](https://github.com/Gunivers/Bookshelf/pull/222)) Add randomized fill function `#bs.block:fill_random`.
+- ([1](https://github.com/Gunivers/Bookshelf/issues/1)) ([220](https://github.com/Gunivers/Bookshelf/issues/220)) Add `bs.random` module for random distributions and noise generation.
```
theogiraudet, line 9: It could be interesting to provide more details using sub-bullets.

docs/modules/random.md
```diff
@@ -0,0 +1,400 @@
...
+  **Storage `bs:out random.choose`**: {nbt}`any` The randomly selected value.
+
+  **Return**: The index of the chosen value.
+```
+
+*Pick a random fruit from the list:*
+
+```mcfunction
+# Populate list with values
+data modify storage bs:in random.choose.options set value ["Apples", "Bananas", "Strawberries", "Blueberries", "Mango", "Watermelon", "Honeydew Melon"]
```
theogiraudet, line 45: Some fruits are singular, others are not.

docs/modules/random.md
```diff
@@ -0,0 +1,400 @@
...
+function #bs.random:choose
+
+# Display the result
+tellraw @a [{"text":"Value: ","color":"dark_gray"},{"nbt":"random.choose","storage":"bs:out","color":"gold"}]
+```
+
+> **Credits**: SBtree
+
+
+### Random distributions
```
theogiraudet, line 57: Is it possible to add an image to see the difference between the different distributions?
And I think it could be interesting to add a page/video that explains the distribution for each.

docs/modules/random.md
```diff
@@ -0,0 +1,400 @@
...
+# Display the result
+tellraw @a [{"text": "Number: ", "color": "dark_gray"},{"score":{"name":"$random.uniform", "objective": "bs.out"}, "color": "gold"}]
+```
+
+::::
+::::{tab-item} Binomial
+
+```{function} #bs.random:binomial {trials:<trials>,probability:<probability>}
+
+Generates a random number with a binomial distribution using the specified `trials` and `probability`.
```
theogiraudet, line 93: What is "trials"?
Maybe some explanation could be interesting.

docs/modules/random.md
```diff
@@ -0,0 +1,400 @@
...
+# Display the result
+tellraw @a [{"text": "Number: ", "color": "dark_gray"},{"score":{"name":"$random.geometric", "objective": "bs.out"}, "color": "gold"}]
+```
+
+::::
+::::{tab-item} Poisson
+
+```{function} #bs.random:poisson {lambda:<lambda>}
+
+Generates a random number following a Poisson distribution with the expected value `lambda`.
```
theogiraudet, line 150: What is lambda?

theogiraudet (changes requested):
This module is really great and complex.
We want to keep Bookshelf as accessible as possible so I think we need to add more images/references/explanations to help users to understand what these functions do.

theogiraudet:
By the way, it would be nice to have an alternative to `#bs.random:choose` that allows weighed values since Minecraft mainly works with that.

theogiraudet (approved):

PR #259, opened 2024-10-09
✨ Add `bs.block:spawn_block_particles`

theogiraudet:
# Tasks to do before merging
  - [x] I have [tested](https://bookshelf.docs.gunivers.net/en/latest/contribute/debug.html#unit-tests) my contribution


aksiome:
Everything looks fine, i'm just not convinced by the name. All other spawn functions are summoning a new entity which is not really the case here. Maybe something like `emit_block_particles` would be more appropriate?

theogiraudet:
I agree, but the folder is named "spawn" :/

aksiome:
We can just move the function to an other folder. Also I think we should add a very simple test to just see if the run function is working (in case mojang change the particle command for example)

theogiraudet:
For me, renaming the folder to "produce" would be a better idea. But this is a breaking change…
+1 for the test.

aksiome:
You can rename it it's not breaking change since functions are tagged and the internals are not supposed to be used directly. Im completly fine with this change

theogiraudet:
Oh this is true, my bad. Let's do that, this name is better for new features.

theogiraudet:
@aksiome I let you review the new version.

aksiome (approved):

PR #268, opened 2024-10-19
🐛 Fix the distribute script workflow

aksiome:
# Tasks to do before merging
- [ ] I have added my contribution to the [changelog](https://github.com/Gunivers/Bookshelf/blob/master/docs/CHANGELOG.md)
  - [ ] I have documented my contribution (`/docs`)
  - [ ] I have added or updated the [header](https://bookshelf.docs.gunivers.net/en/latest/contribute/file-format.html) of the features' root function I contribute
  - [ ] I have [tested](https://bookshelf.docs.gunivers.net/en/latest/contribute/debug.html#unit-tests) my contribution

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
           echo "BUILD_PATH=$(python -c 'import definitions; print(definitions.BUILD_PATH)')" >> $GITHUB_ENV
 
       - name: Package datapacks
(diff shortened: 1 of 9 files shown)


aksiome:
This PR also enables the definition of module dependencies using the module.json file. Some dependencies were missed due to the use of macros or because direct functions were called instead of using function tags. While we should be able to address the direct function issue, it's unclear how to resolve the macro-related ones.

Issue #191, opened 2024-05-11
`bs.string`: A module to manage strings


theogiraudet:
New facility functions to manipulate string:
- [ ] To char array
- [ ] Index of
- [ ] Split according to a char

Darukshock:
I will code this

Darukshock:
Should IndexOf that a char as an input or a string ?
Should it output the first index or an array if multiple are present ?

aure31:
I can take this module

aure31:
the different functions offered by the string module

#bs.string:index_of
[inputs]
str: [str] the string
letter: [char] the letter to find
index : [int]the nth occurance index

[output]
bs.out string.index

[use]:
retrieves the index of the first recurrence of the character

#bs.string:append
[inputs]
str1 : [str]
str2 : [str]

[output]
bs:out string.out

#bs.string:lenth
[inputs]
str : [str]

[output]
bs.out string.lenth

#bs.string:lenth
[inputs]
str : [str]

[output]
bs.out string.lenth

#bs.string:sub_string
[inputs]
str : [str]
start : [int]
end : [int]

[output]
bs:out string.out

#bs.string:lower
[inputs]
str : [str]

[output]
bs:out string.out

#bs.string:upper
[inputs]
str : [str]

[output]
bs:out string.out

#bs.string:CharAt
[inputs]
str : [str]
index : [int]

[output]
bs:out string.out

#bs.string:CharAt
[inputs]
str : [str]
index : [int]

[output]
bs:out string.out

#bs.string:to_arrays
[inputs]
str : [str]

[output]
bs:out string.array [list]

#bs.string:split
[inputs]
str : [str]
caracter : [char]

[output]
bs:out string.array [list]

#bs.string:replace
[inputs]
str : [str]
char : [char]
new_char : [char]

[output]
bs:out string.out [str]

#bs.string:replace_at
[inputs]
str : [str]
index : [int]
new_char : [char]

[output]
bs:out string.out [str]


here's a non-exhaustive list of the different functions what do you think?

theogiraudet:
Hey @aure31 have you found time to start the module?
if you're still up for it, do you think we'll be able to release it in the next iteration (no date set yet)?

aure31:
I start a little bit but the module it self is not too hard to make so I think i can finish it for the next iteration

theogiraudet:
> I start a little bit but the module it self is not too hard to make so I think i can finish it for the next iteration

[...]

PR #283, opened 2024-11-30
✨ feat: add a new `bs.string` module

aure31:
# Tasks to do before merging
  - [x] I have [tested](https://bookshelf.docs.gunivers.net/en/latest/contribute/debug.html#unit-tests) my contribution


aure31:
close #191

modules/bs.string/data/bs.string/function/concat/dispatch.mcfunction
aksiome, line 1: Im not sure you understood completly what i was going for with this. Using the current implementation there is no benefit from the macro cache since the function is called with any length. The whole point was to create small chunks of 8 to avoid cache miss. Also I was maybe a bit wrong and it's better to scale this up to a cache of 8 by 8. Which means going up to 64 before using powers of 2. If you're not sure about it don't worry to much i already have a pretty good idea of the implementation so i can take care of it

modules/bs.string/data/bs.string/function/lower_case/fast/loop.mcfunction
aksiome, line 1: Im not sure we need a fast and slow behavior. This come from the fact that the table is a list and you use filtering. But having one way tables (`{a:A,b:B}`) for uppercase to lowercase and lowercase to uppercase would just simplify all this and actually bring even more performance since they are simple hashmap with no filtering

modules/bs.string/data/bs.string/function/upper_case/fast/loop.mcfunction
aksiome, line 1: same as for lowercase. Also i think upper and lower are clear enough, no need for lower_case and upper_case

aksiome (commented):
Hey really quick review before this gets too big. I have'nt reviewed everything but it's a start

meta/versions.json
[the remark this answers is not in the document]
aure31: normaly resolved I think it's was during the merge or when I try to update the meta my fault sorry

modules/bs.string/data/bs.string/function/concat/dispatch.mcfunction
aure31: I’ve made some changes, but I’m not sure if they align with what you had in mind. If that’s not the case, could you please clarify? I’d really like to understand your perspective.

aure31: So if I understand right, the minecraft cache work with 8 values for each file with the macro so all the value between 0 and 64 here. Is that right ? I would like to make a call to explain me how the cache system work if you can.

aksiome: Yeah minecraft can cache up to 8 different macro functions. For example with these inputs you'll have 3 functions in cache:
- `{a:1,b:0}`
- `{a:0,b:1}`
- `{a:1,b:1}` even if a:1 already exist b is different. It works using the full set of inputs not just a single value

That's also why sometimes it's better to split these macro inputs between multiple function by using a function tag as we do in some modules. This way you get a bit of extra cache.

After 8 different functions are stored in cache the oldest gets deleted to let space for new ones. Also a function that is in cache will perform the same as a function with no macro.

modules/bs.string/data/bs.string/function/find/find.mcfunction
```diff
@@ -0,0 +1,35 @@
...
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+$scoreboard players set #bs.string.occurrence bs.data $(occurrence)
```
aksiome, line 16: no need to use bs.data can be bs.ctx if the function does not "leak" into user territory (with a callback) or with other functions from other modules

modules/bs.string/data/bs.string/function/find/find.mcfunction
```diff
@@ -0,0 +1,35 @@
...
+data modify storage bs:out string.find set value []
+scoreboard players set #c bs.ctx 0
+execute store result storage bs:ctx x int 1 run scoreboard players set #i bs.ctx 0
+data modify storage bs:ctx _ set from storage bs:in string.find
+execute store result score #l bs.ctx run data get storage bs:in string.find.str
```
aksiome, line 21: same as before to get the len

modules/bs.string/data/bs.string/function/find/inverse/inverse.mcfunction
```diff
@@ -0,0 +1,26 @@
...
+# This source code is subject to the terms of the Mozilla Public License, v. 2.0.
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+say inverse
```
aksiome, line 15: remaining say

modules/bs.string/data/bs.string/function/find/inverse/up.mcfunction
aksiome, line 1: missing header

modules/bs.string/data/bs.string/function/find/precompute.mcfunction
```diff
@@ -0,0 +1,21 @@
...
+#
+# This source code is subject to the terms of the Mozilla Public License, v. 2.0.
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
```
aksiome, line 14: missing new line

modules/bs.string/templates/concat.jinja
aksiome, line 1: The 2 templates are outputting an extra line after the header, also adding a comment like this `# This file was automatically generated, do not edit it` at the start might be a good thing, see the templates from the block module for reference

aksiome (commented):
Ok i'm really nitpicking here, i mainly reviewed the style and not the logic. For the logic I'll need more time to run the tests and all. Also there are some remaining say and tellraw, i'm sure it's because it's not completly finished and ready for the last review, but i commented some of them (not all). A small cleanup cannot hurt i guess 👍

modules/bs.string/data/bs.string/function/find/find.mcfunction
aure31: with the if data ? because it's doesn't work for me it's juste store the success and not the len

modules/bs.string/data/bs.string/function/find/inverse/up.mcfunction
aure31: for the reversed I maybe will delete it. I d'ont what you think ? because I need to do a different precompute for it

modules/bs.string/data/bs.string/function/find/find.mcfunction
aksiome: yeah with if data it works but you need [] at the end `if data storage path to.array[]`

modules/bs.string/data/bs.string/function/find/inverse/up.mcfunction
aksiome: as you want, but yeah maybe a single direction is enough

modules/bs.string/data/bs.string/function/find/find.mcfunction
aure31: but it's a string not a list

docs/modules/string.md
```diff
@@ -0,0 +1,100 @@
...
+```{function} #bs.string:upper {type:<mode>}
+
+Converts the text to uppercase
+
+:Inputs:
+  **Storage `bs:in string.upper.str`**: {nbt}`string` The string to convert to uppercase
+
+  **Function Macro**:
+  :::{treeview}
+  - {nbt}`compound` Arguments
```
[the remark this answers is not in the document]
aure31: with what I replace it ?

modules/bs.string/data/bs.string/function/find/find.mcfunction
aksiome: Yeah sorry, it's only for concat. Did too much python haha 😅

docs/modules/string.md
```diff
@@ -0,0 +1,100 @@
+# 🔠 String
+
+**`#bs.string:help`**
```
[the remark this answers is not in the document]
aure31: how you find the image for the module with a black and white variant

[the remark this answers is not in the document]
aksiome: Either have a png with transparent background, or make 2 versions. If you don't really feel like doing some image design we'll try to do it (either Theo or me)

modules/bs.string/data/bs.string/function/find/normal/normal.mcfunction
```diff
@@ -0,0 +1,26 @@
...
+
+$data modify storage bs:ctx _.test set string storage bs:ctx _.str 0 $(y)
+data modify storage bs:ctx _.ltr set string storage bs:ctx _.test -1
+execute store success score #t bs.ctx run data modify storage bs:ctx _.test set from storage bs:in string.find.needle
+
+execute if score #t bs.ctx matches 0 run data modify storage bs:out string.find append from storage bs:ctx x
+
+data modify storage bs:ctx z set from storage bs:ctx y
+function bs.string:find/match_patern with storage bs:ctx _
+execute if score #t bs.ctx matches 0 run return run function bs.string:find/normal/skip with storage bs:ctx
```
aksiome, line 24: Not really needed since we're calling skip either way right?

modules/bs.string/data/bs.string/function/find/normal/normal.mcfunction
```diff
@@ -0,0 +1,26 @@
...
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+$data modify storage bs:ctx _.test set string storage bs:ctx _.str 0 $(y)
+data modify storage bs:ctx _.ltr set string storage bs:ctx _.test -1
+execute store success score #t bs.ctx run data modify storage bs:ctx _.test set from storage bs:in string.find.needle
+
+execute if score #t bs.ctx matches 0 run data modify storage bs:out string.find append from storage bs:ctx x
+
+data modify storage bs:ctx z set from storage bs:ctx y
```
aksiome, line 22: no need to copy this value, directly use y instead of z inside skip

modules/bs.string/data/bs.string/function/find/normal/skip.mcfunction
```diff
@@ -0,0 +1,22 @@
...
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+$execute store result storage bs:ctx x int 1 run scoreboard players add #i bs.ctx $(z)
```
aksiome, line 16: As said in find, it's better to only work with score and only copy to out storage when something is found

modules/bs.string/data/bs.string/function/find/count/up.mcfunction
```diff
@@ -0,0 +1,21 @@
...
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+data modify storage bs:out string.find append from storage bs:ctx x
```
aksiome, line 16: As said after in find avoid using x and use the score instead

modules/bs.string/data/bs.string/function/find/count/count.mcfunction
```diff
@@ -0,0 +1,24 @@
...
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+$data modify storage bs:ctx _.test set string storage bs:ctx _.str 0 $(y)
+data modify storage bs:ctx _.ltr set string storage bs:ctx _.test -1
+execute store success score #t bs.ctx run data modify storage bs:ctx _.test set from storage bs:in string.find.needle
+
+data modify storage bs:ctx z set from storage bs:ctx y
```
aksiome, line 20: not needed directly use y directly in match_pattern instead

modules/bs.string/data/bs.string/function/find/count/skip.mcfunction
```diff
@@ -0,0 +1,22 @@
...
+# This source code is subject to the terms of the Mozilla Public License, v. 2.0.
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
```
aksiome, line 15: Same as for normal/skip

modules/bs.string/data/bs.string/function/split/split.mcfunction
```diff
@@ -0,0 +1,46 @@
...
+#
+# This source code is subject to the terms of the Mozilla Public License, v. 2.0.
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
```
aksiome, line 14: missing new line. Also im lazy and i think a lot of code is shared with find, so i'll just say everything for find applies here

modules/bs.string/data/bs.string/function/replace/replace.mcfunction
```diff
@@ -0,0 +1,46 @@
...
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+#reset
```
aksiome, line 16: comment style. Also im lazy so i'll just say that everything from find applies here

aksiome (changes requested):
Globally i think we should avoid mixing storage and macro. This is because it's really hard to tell if some parameters are to be used dynamically or not. For example occurence may be used with a constant value but some people might want to use it dynamically and it adds extra complexity. I think we should only use storage for the whole module. Also some tests are failing, and i think some function could be changed a bit. Here are my propositions:

- find should be renamed find_all (still returns a list) and we should have a simple find that returns only the first occurence as an int from the function. This mean that the command can easily be used with if function.
- insert can be reworked easily into something like replace_range / substr_replace which does the same and more with a similar implementation (see the details after)

modules/bs.string/data/bs.string/function/insert/insert.mcfunction
```diff
@@ -0,0 +1,22 @@
...
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+$data modify storage bs:ctx _.s.3 set string storage bs:in string.insert.str 0 $(index)
+data modify storage bs:ctx _.s.2 set string storage bs:in string.insert.needle
+$data modify storage bs:ctx _.s.1 set string storage bs:in string.insert.str $(index)
+
```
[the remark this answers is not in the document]
aure31: yes but it's relly less optimise because he need to check for the size who is always 3 so we lost performance for nothing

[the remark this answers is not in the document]
aksiome: Sorry if i was unclear, i was not talking about using the concat function, i was saying we should copy the content of the bs.string:concat/combine/2c function into a dedicated concat function inside the insert folder

modules/bs.string/README.md
```diff
@@ -0,0 +1,9 @@
+# 🔠 Bookshelf String
+
+This module allows you to easily manage string in Minecraft.
```
theogiraudet, line 3: This module allows you to easily manage strings in Minecraft.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+
+Manage strings (text), allowing easy transformation and searching within the text.
+
+```{admonition} Technical Limitation
+:class: warning
+
+Due to the way string are handled, we don't support using `"` you need to use the `\"` instead.
+
+```
+
```
theogiraudet, line 13: It could be interesting to add as a global limitation/warning the range of supported characters. I suppose you cannot lowercase an ancient-Greek letter for instance.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+Convert text to uppercase.
+
+:Inputs:
+  **Storage `bs:in string.upper.str`**: {nbt}`string` The string to convert to uppercase.
+
+:Outputs:
+  **Storage `bs:out string.upper`**: {nbt}`string` The uppercase string.
+```
+
+_Convert `hello world` to uppercase:_
```
theogiraudet, line 63: The example must help the user to understand the feature, here you are not showing the expecting result so we cannot understand it if we don't know what "uppercase" means.
I suggest to modify this line by "Convert `hello world` to its uppercase version `HELLO WORLD`:" or something like this.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+Convert text to lowercase.
+
+:Inputs:
+  **Storage `bs:in string.lower.str`**: {nbt}`string` The string to convert to lowercase.
+
+:Outputs:
+  **Storage `bs:out string.lower`**: {nbt}`string` The lowercase string.
+```
+
+_Convert `HELLO WORLD` to lowercase:_
```
theogiraudet, line 85: Same here

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+
+Search for a substring within a string. Locate all occurrences or a specific number of matches.
+
+:Inputs:
+  **Storage `bs:in string.find`**:
+  :::{treeview}
+  - {nbt}`compound` Find data
+    - {nbt}`string` **str**: String to search in.
+    - {nbt}`string` **substr**: String to find.
+    - {nbt}`int` **occurrence**: Number of occurrences to find (0 for all).
```
theogiraudet, line 112: Semantically, I'm not a big fan of "zero" as "all occurrences" because 0 means to search for no occurences, which even if this is a trivial case, it is semantically correct.
According to me, `-1` may be better because "-1 occurences" means nothing. This number is often use to say "no limit".
Do you have a reason for using 0 (an example in a language, optimization reason)?
If not, I suggest to use `-1` instead.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+
+  **Return**: Number of occurrences found.
+```
+
+_Find `world` in `hello world`:_
+
+```mcfunction
+data modify storage bs:in string.find set value {str:"hello world",substr:"world",occurrence:0}
+function #bs.string:find
+tellraw @a [{"text":"Found at indices: \""},{"storage":"bs:out","nbt":"string.find"},{"text":"\""}]
```
theogiraudet, line 126: Put in comment the two expected results. We should be able to read the results without having to execute the command.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+
+> **Credits**: Aure31
+
+---
+
+### Replace
+
+```{function} #bs.string:replace
+
+Replace occurrences of a substring with new text.
```
theogiraudet, line 137: "Replace occurrences of a substring by a new text."

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+
+:Inputs:
+  **Storage `bs:in string.replace`**:
+  :::{treeview}
+  - {nbt}`compound` Replace data
+    - {nbt}`string` **str**: Base string for replacements.
+    - {nbt}`string` **old**: Substring to replace.
+    - {nbt}`string` **new**: Replacement string.
+    - {nbt}`int` **maxreplace**: Maximum replacements. (0 for unlimited)
```
theogiraudet, line 146: Same comment than before.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+:Outputs:
+  **Storage `bs:out string.replace_range`**: {nbt}`string` The modified string with the specified range replaced.
+```
+
+_replace the space in `hello world` by ` beautiful `:_
+
+```mcfunction
+data modify storage bs:in string.replace_range set value {str:"hello world",substr:" beautiful ",start:5,end:6}
+function #bs.string:replace_range
+tellraw @a [{"text":"The new string is \""},{"storage":"bs:out","nbt":"string.replace_range"},{"text":"\""}]
```
theogiraudet, line 190: Same than previous comment.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+:Outputs:
+  **Storage `bs:out string.reverse`**: {nbt}`string` The string with all characters in reverse order.
+```
+
+_reverse `hello world`:_
+
+```mcfunction
+data modify storage bs:in string.reverse.str set value "hello world"
+function #bs.string:reverse
+tellraw @a [{"text":"The reversed string is \""},{"storage":"bs:out","nbt":"string.reverse"},{"text":"\""}]
```
theogiraudet, line 220: Same than previous comment.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+
+Divide a string into smaller parts using a specified separator, creating a list of substrings.
+
+:Inputs:
+  **Storage `bs:in string.split`**:
+  :::{treeview}
+  - {nbt}`compound` Split data
+    - {nbt}`string` **str**: The string to be split into parts.
+    - {nbt}`string` **separator**: The character or string that marks where to split.
+    - {nbt}`int` **maxsplit**: Maximum number of splits to perform (0 or negative for unlimited).
```
theogiraudet, line 239: Same than previous comment.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+:Outputs:
+  **Storage `bs:out string.split`**: {nbt}`list` A list containing all the split parts of the original string.
+```
+
+_split `hello world` by `" "`:_
+
+```mcfunction
+data modify storage bs:in string.split set value {str:"hello world",separator:" ",maxsplit:0}
+function #bs.string:split
+tellraw @a [{"text":"The list of strings is \""},{"storage":"bs:out","nbt":"string.split"},{"text":"\""}]
```
theogiraudet, line 251: Same than previous comment.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+---
+
+### Transform Type
+
+:::::{tab-set}
+::::{tab-item} To List
+
+```{function} #bs.string:to_list
+
+Convert a string into a list where each character become a separate element.
```
theogiraudet, line 265: "Convert a string into a list where each character become**s** a separate element."

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+:Outputs:
+  **Storage `bs:out string.to_list`**: {nbt}`list` A list containing each character as a separate element.
+```
+
+_transform `hello world` into a list:_
+
+```mcfunction
+data modify storage bs:in string.to_list.str set value "hello world"
+function #bs.string:to_list
+tellraw @a [{"text":"The list of characters is \""},{"storage":"bs:out","nbt":"string.to_list"},{"text":"\""}]
```
theogiraudet, line 279: Same than previous comment.

docs/modules/string.md
```diff
@@ -0,0 +1,329 @@
...
+```
+
+::::
+::::{tab-item} To String
+
+```{function} #bs.string:to_string
+
+Convert any value into it's string representation.
```
theogiraudet, line 287: "Convert any value into it**s** string representation."

theogiraudet (changes requested):

docs/modules/string.md
aure31: yes you can it's support all the utf8 table

theogiraudet: So it can be interesting to indicate that the module support all the UTF-8 table ;)

aure31: do I put it into a admiton type hint after the technicall limitation ?
or somewhere else ?

theogiraudet: After technical limitation I think? @aksiome

aksiome: If we want to be really rigorous i think that the `"` limitation should be below every function that is concerned but not at the top. This is because it will actually just not work at all for uppercase and lowercase, since `\"` will be read as 2 characters (which is something we can easily fix). Also in find `\"` will also not be correctly read in the skip table so it's not gonna work at all. The only real limitation with this is when concatenating, and the problem is that find (and all other similar functions) use concatenation. This is why i think we should truly precise in each function what is the expected result when working with `"`, either we cant use them at all or we need to escape them

aksiome: And so to actually answer the question i think the precision about supported char should be inside the convert case section only imo. Because everything else is just minecraft

aure31: I have change it but my phrase aren't good so tell me what I need to put

docs/modules/string.md
```diff
@@ -0,0 +1,424 @@
...
+```mcfunction
+data modify storage bs:in string.concat.list set value ["Hello ","World"]
+function #bs.string:concat
+tellraw @a [{"text":"The merged string is \""},{"storage":"bs:out","nbt":"string.concat"},{"text":"\""}]
+```
+
+```{admonition} Technical Limitation
+:class: warning
+
+Due to the way string are handled, we don't support using `"` you need to use the `\"` instead.
```
theogiraudet, line 39: "If you want to use double quotes (`"`) in a string, you need to escape them: `"\""`."
Is it also true for single quotes `'`? If yes, do we need to escape double quotes when we use single quotes and vice-versa?

docs/modules/string.md
```diff
@@ -0,0 +1,424 @@
...
+tellraw @a [{"text":"The lowercase string is \""},{"storage":"bs:out","nbt":"string.lower"},{"text":"\""}]
+```
+
+::::
+:::::
+
+```{admonition} Support
+:class: hint
+
+this function actually support all UTF-8 characters so you can use it with any language.
```
theogiraudet, line 99: "This function actually supports all UTF-8 characters."
I don't think we need to specify we "can use it with any language".

theogiraudet (changes requested):

docs/modules/string.md
aure31: No only double quote

aksiome:
Since we're gonna do some more changes on this module in #375. I propose to merge it as it is, and we'll improve the phrasing in #375 even if it's not super clean. I'll wait this evening in case someone is against this

aksiome (approved):

PR #310, opened 2025-01-25
Add bs.block:play_block_sound

### LINKED ISSUE #279: `bs.block:play_sound`: Play the sound of a block
This feature should take the type of sound we want to play as a parameter: `break`, `hit`, `place`, or `step`.
Else, the idea is similar to `emit_block_particle` and other "virtual block" to "command" features.

theogiraudet:
### Description
Add a new feature to play a sound of a getted block with bs.block module..

### Related Issues
Closes #279

### Additional Notes
/

- [x] I have updated the [changelog](https://github.com/mcbookshelf/Bookshelf/blob/master/docs/CHANGELOG.md) to reflect my contribution.
  - [ ] I have thoroughly tested my changes. See [testing guidelines](https://docs.mcbookshelf.dev/en/latest/contribute/debug.html#unit-tests).


modules/bs.block/data/bs.block/function/produce/block_sound/play_block_sound.mcfunction
```diff
@@ -0,0 +1,29 @@
...
+# If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+data modify storage bs:ctx _ set value {}
```
aksiome, line 16: could be simplified to data modify storage bs:ctx _ set from storage bs:in block.play_block_sound

aksiome: Also looking at that, do we really need to pass the whole sounds map i feel like the user can do that better and it avoid the use of a macro, and actually also no need to log since the user is now responsible for checking if the storage contains the sound. Also it's as simple to use as your version:
```mcfunction
data modify storage bs:in block.play_block_sound set value { source: "block", targets: "@s", pos: "~ ~ ~", volume: 1, pitch: 1, min_volume: 0 }
data modify storage bs:in block.play_block_sound.sound set from storage bs:out block.sounds.break
```

aksiome (changes requested):

modules/bs.block/data/bs.block/function/produce/block_sound/play_block_sound.mcfunction
theogiraudet: Indeed we can. Even if the idea was, like with other producing functions, to just be able to drop the whole block data inside the input.

aksiome: Yeah i can see the idea behind that but i'm not sure it's worth it considering that:
- the alternative is as easy to use
- has better performances
- is actually less code on our side (so less things to worry about)

theogiraudet: If you think this is better I can change it. But currently, I think we consider that users never have to know what is the structure of the virtual blocks. Here, we are breaking this expectation. Are you good with that?

aksiome: Personally i don't mind, also the fact that the user should not know what is the structure is not really true since we explain the format on the doc (what we dont explain is the extra key `_` that is used to manipulate the output). So the most important part is that it's readonly and that the user should not modify directly the output

aksiome: Also i think it'll be nice to add to the doc the different sound types that the module returns

docs/modules/block.md
```diff
@@ -893,6 +905,46 @@ data modify storage bs:in block.emit_block_particle merge value { delta: "0 0 0"
...
+
+Play a block sound of the given block.
+
+:Inputs:
+  **Execution `at <entity>` or `positioned <x> <y> <z>`**: Position where the sound will be played.
+
+  **Storage `bs:in block.play_block_sound`**:
+  :::{treeview}
+  - {nbt}`compound` Block sound data
+    - {nbt}`string` **sound**: The sound to play. Usually took from the `sounds` property of the virtual block (cf [Get](#get) functions).
```
aksiome, line 921: The get reference causes a fatal sphinx error. It may be related to the treeview plugin but i guess the easiest solution is to remove the reference for now

aksiome (changes requested):

aksiome (approved):

PR #373, opened 2025-03-22
📝 docs: setblock now supports `strict` mode

### LINKED ISSUE #363: Setblock now supports `strict` mode, we need to document that

theogiraudet:
### Description
Since Minecraft 1.21.5, the setblock has now a new mode: `strict`.
As we use the setblock in some `bs.block` features, we should update the documentation to indicate this new mode.

### Related Issues
Fixes #363

- [x] I have updated the [changelog](https://github.com/mcbookshelf/bookshelf/blob/master/docs/CHANGELOG.md) to reflect my contribution.
  - [ ] I have thoroughly tested my changes. See [testing guidelines](https://docs.mcbookshelf.dev/en/latest/contribute/debug.html#unit-tests).


docs/_templates/changelog/v3.0.0.md
```diff
@@ -30,6 +30,7 @@
 - <abbr title="New Features">✨</abbr> **[#279](https://github.com/mcbookshelf/Bookshelf/issues/279)** - Added the `#bs.block:play_block_sound` function for playing block sounds.
 - <abbr title="New Features">✨</abbr> **[#287](https://github.com/mcbookshelf/Bookshelf/issues/287)** - Updated `replace_type` to return whether a type was found and replaced.
 - <abbr title="Bug Fix">🐛</abbr> **[#320](https://github.com/mcbookshelf/Bookshelf/issues/320)** - Fixed functions that were unusable outside the Overworld.
+- <abbr title="Enhancements">⚡</abbr> **[#363](https://github.com/mcbookshelf/bookshelf/issues/363)** - `fill_block`, `fill_type`, `fill_random`, and `set_block`, `set_type` now accept `strict` for the `mode` parameter to specify whether the operation should trigger a block or shape updates.
```
aksiome, line 33: Since each changelog line starts with a verb in the past tense, we should maintain the same format (also should trigger a block or shape updates, i know this is mc changelog but it's super confusing for the shape part because it will actually replace blockstates and use the default one, i think we should only keep: accept new option strict, which places blocks as-is without block updates.)

aksiome (changes requested):

aksiome (approved):

PR #398, opened 2025-04-05
🐛 fix(bs.health): concurrency issue with external damage sources

### LINKED ISSUE #395: Tick issue when healing the player
### Module

health

### Feature

*_health

### Version

3.0.0

### Expected Behavior

When modifying health (e.g., healing or setting it), the changes should take effect immediately (at least from the user standpoint). Any subsequent commands, such as dealing damage right after healing, should work as expected.

### Actual Behavior

When using Bookshelf to change health, any health-related changes made after the healing command through other means than Bookshelf (e.g., using the damage command) are discarded. This means the player’s health does not reflect those later changes, as if the Bookshelf system overwrites them.

### Possible Fix

In version 3.0.0 (1.21.5), a visual fix changed the way health was updated: the attribute modifier is now applied after the instant_health effect, to bring the player's health down to the desired value. This is done using a score that represents the health to set.

Instead, we should use a score that represents how much health to remove, not set. This would avoid overwriting newer changes made after healing. However, we need to be careful with max_health, since it may change in between and affect how much should be removed.

### Steps to Reproduce

_No response_

### LINKED ISSUE #396: Inconsistency in `get_max_health` behavior
### Suggestion or Inquiry

Currently, the `get_max_health` function is inconsistent with `add_max_health` and `set_max_health`. While `add_max_health` and `set_max_health` modify the base attribute, `get_max_health` retrieves the total value, including all modifiers. Additionally, `get_max_health` was introduced as a counterpart to `get_health`, but it is somewhat redundant. Users can achieve the same result by directly using: `attribute @s minecraft:max_health get <scale>`

aksiome:
### Related Issues
- closes #395
- closes #396


theogiraudet (changes requested):
Changelog?

PR #408, opened 2025-04-07
⚙️ ci: secure the release pipeline and improve issue templates

### LINKED ISSUE #400: Add beet cache to github workflows
### Suggestion or Inquiry

Use the beet cache in workflows to speed up things

### LINKED ISSUE #406: Fix the release pipeline
### Suggestion or Inquiry

Currently the workflow needs permission to push to the master branch in order to update versions metadata and the documentation switcher. We should rethink this process

aksiome:
### Related Issues
- closes #406
- closes #400


aksiome:
With this PR, we currently have the following issue types:

- Bugs
- Features (new modules or new functionality)
- Tasks (with various categories)
- Other

@theogiraudet, do you think we should add a dedicated type for feature improvements or enhancements? Or should those be reported under "Task" or "Feature"?

theogiraudet:
Hum I don't have a strong opinion about. I think the question is: do we have specific information to ask for an improvement?

aksiome:
The question is that we have currently 3 types (not label but issue types):
- 🛠 Task
- 🐛 Bug
- ✨ Feature

I updated the issue templates to have 4 templates, a free one, and one for each type. If the user wants to request an improvement (like https://github.com/mcbookshelf/bookshelf/issues/409) which type should be used? Should we create a new type for that or use one that already exists?

theogiraudet:
Yeah I understand, but my point is that I don't think having several templates with the same questions is useful. So if we have different information to get, I suppose it is interesting to have more templates, else may change the name of the existing templates to be more general.

aksiome:
Templates aren't just for guiding questions—they also allow you to automatically assign an issue type. Personally, I don't mind having duplicate questions if the template helps categorize them with a different issue type.

theogiraudet:
In all cases, we have to manually add module tags, so I don't see the utility having redundant templates.
If we have other information to ask, then why not?

aksiome:
Yeah, we’ll always need to update labels, but not having to set types manually and having things pre-sorted is definitely a nice advantage. Also, I think this PR is already ready, I was just curious about your opinion. The question was more about how we categorize things. For example, should a performance improvement be categorized as a Feature or a Task?

theogiraudet (approved):

PR #475, opened 2025-07-11
✨ feat(bs.environment): add new daylight features

### LINKED ISSUE #14: Rotation to moon/sun
Put the rotation of an armor stand/other so that it looks/points at the sun/moon

theogiraudet:
### Description
Add six new function features in bs.environment:
- Get moon/sun position in degrees for a given time
- Get moon/sun position at the current time
- Orient the executing entity to look at the sun/moon

Also add twelve new predicates relative to daylight cycle:
- is day/sunset/night/sunrise time?
- are villagers sleeping/socializing/working?
- can bed be used?
- can undeads burn?
- is creakings spawn period?
- is monsters spawn period?
-  are bees sleeping?

### Related Issues
closes #14 


modules/bs.environment/data/bs.environment/function/__load__.mcfunction
```diff
@@ -20,4 +20,18 @@ scoreboard objectives add bs.const dummy [{text:"BS ",color:"dark_gray"},{text:"
...
+scoreboard players set 500 bs.const 500
+scoreboard players set 360 bs.const 360
+scoreboard players set 250 bs.const 250
+scoreboard players set 180 bs.const 180
+scoreboard players set 10 bs.const 10
+scoreboard players set 8 bs.const 8
+scoreboard players set 2 bs.const 2
+scoreboard players set -1 bs.const -1
+
+data modify storage bs:data environment.moon_phases set value ["full_moon", "waning_gibbous", "third_quarter", "waning_crescent", "new_moon", "waxing_crescent", "first_quarter", "waxing_gibbous"]
```
aksiome, line 37: Should use const instead of data

modules/bs.environment/data/bs.environment/function/celestial/get_sun_angle.mcfunction
```diff
@@ -0,0 +1,86 @@
...
+# In the original algorithm, d is multiplied by PI to obtain the angle in radians for the cosine function.
+# The cosine function from Bookshelf expects the angle in degrees, so we need to convert it to degrees instead of radians.
+scoreboard players operation #e bs.ctx *= 180 bs.const
+
+# Call cosine function with precision constraint
+# Prepare input: convert from scale 10³ to 10², since this is the precision of the Bookshelf's cosine function
+scoreboard players operation $math.cos.x bs.in = #e bs.ctx
+scoreboard players operation $math.cos.x bs.in /= 10 bs.const
+# Call the function
+function #bs.math:cos
```
aksiome, line 60: I feel like having a dep on math here is not worth it. We should inline the cos since it's a quite a simple function, also it's gonna be easier to stay on ctx and avoid copying scores around

modules/bs.environment/data/bs.environment/function/celestial/look_at_moon.mcfunction
```diff
@@ -0,0 +1,20 @@
...
+#
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+execute store result storage bs:ctx y float 0.001 run function bs.environment:celestial/get_moon_angle
+execute store result score #t bs.ctx run time query daytime
```
aksiome, line 17: Maybe this command can be avoided by reusing the one in get angle. But that's not super clean so unsure

modules/bs.environment/data/bs.environment/function/celestial/look_at_sun.mcfunction
```diff
@@ -0,0 +1,20 @@
...
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+execute store result storage bs:ctx y float 0.001 run function bs.environment:celestial/get_sun_angle
+execute store result score #t bs.ctx run time query daytime
+execute if score #t bs.ctx matches 6000..18000 run data modify storage bs:ctx x set value 90
```
aksiome, line 18: Since it's only 2 possible states i think it's better in this case to have 2 rotate function and a single macro param.
Then we have a single if score with a return run function and the other rotate function just after

modules/bs.environment/data/bs.environment/tags/function/get_moon_angle.json
```diff
@@ -0,0 +1,20 @@
+{
+  "__bookshelf__": {
+    "feature": true,
+    "documentation": "https://docs.mcbookshelf.dev/en/latest/modules/environment.html#get-moon-angle",
```
aksiome, line 4: This anchor is i think wrong and will direct to nothing should be celestial-movement (a valid title in the doc), same for all others

aksiome (changes requested):
Maybe we can add tests for angle related functions with precomputed values? Anyway it's not blocking for me but worth considering

docs/modules/environment.md
```diff
@@ -84,6 +84,135 @@ function #bs.environment:get_temperature {scale:1000}
...
+:::::{tab-set}
+::::{tab-item} Get Sun Angle
+
+```{function} #bs.environment:get_sun_angle
+
+Get the sun's angle on the Y axis relative to the horizon, in degrees.
+Also works when the doDaylightCycle is set to false.
+
+:Outputs:
+  **Return | Score `$environment.get_sun_angle bs.out`**: Sun's angle in degrees (scaled by 1000).
```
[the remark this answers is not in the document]
theogiraudet: I agree, but cosine is restricted to 10² in input and 10³ in output so I think we cannot change the scaling as we want.

theogiraudet:
I tried, but I met some issues by changing time and checking the output of my function with Packtest.

docs/modules/environment.md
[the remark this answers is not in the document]
aksiome: I dont think the scaling we use internally matter maybe just use a storage to store the real result then in the return use:
return run execute store result score output run data get storage value $(scale)

This way if we want to change the accuracy in the future we can still do it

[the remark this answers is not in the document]
theogiraudet: > I dont think the scaling we use internally matter

I can check, but the accuracy we propose is, I believe, bound to the lowest accuracy we use in our computation.
I started to check with an accuracy of 10², but the accuracy was too bad, with an inaccuracy of, sometimes, ±5°.

modules/bs.environment/data/bs.environment/function/celestial/look_at_moon.mcfunction
theogiraudet: I propose to split "get_X_angle" into two different features:

- get_current_X_angle, which will have the same behavior
- get_X_angle, which proposes to specify the daytime.

loot_at_X will basically call the second one to avoid this getting daytime duplication.
What do you think?

modules/bs.environment/data/bs.environment/function/celestial/get_sun_angle.mcfunction
theogiraudet: If you are okay with, I will just duplicate the function and change the input/output storage.

aksiome: I dont really care but i personally think it's better to inline it since we can certainly save a few commands this way. But this can be done in a future update

theogiraudet: cf what I've done.
I think this is similar to what you are talking about, but placed in a different file.

aksiome: Having it inline means one less function call, but if it's easier this way np

theogiraudet:
> Even when adding a delay? Like waiting a few tick or so after changing the time?

My test and the result, as you can see we don't obtain the correct result…
image

modules/bs.environment/data/bs.environment/function/celestial/get_sun_angle.mcfunction
theogiraudet: Yeah, but also less clear, I think. I prefer to keep clarity, especially when we are talking about one function call.

docs/modules/environment.md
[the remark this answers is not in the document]
theogiraudet: Small issue I can see: 
Since we are currently at a scale of 1000, to have the result at 10³ the user must specify "1" instead of "1000".
Two solutions:
- We keep like that and we document.
- I first scale in the storage at 0.001, then I set it again from the storage to the storage with the scale given by the user.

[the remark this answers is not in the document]
theogiraudet: Hum, I tried the last solution but… I lose accuracy…
Basically, I need to first convert to scale 1:
```
execute store result storage bs:out environment.celestial_angle float 0.001 run scoreboard players get $r bs.ctx
```
Then to convert to the required scale:
```
$execute store result storage bs:out environment.celestial_angle float $(z) run data get storage bs:out environment.celestial_angle
```
_As far as I know, we cannot do this through /data command_
For a value of "-12240"  and a scale of 1000, the first command will save `-12.24` but the last one will save `-13000` instead of "-12240"…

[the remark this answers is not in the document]
aksiome: I think it's better to use double (default without the f suffix)
Then to retrieve the data you just have to do as i mentionned earlier and use the scale in the data get:
```mcfunction
$return run data get storage bs:out environment.get_celestial $(scale)
```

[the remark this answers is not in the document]
aksiome: I should have been clearer that the scale only impact the returned or the score value if we use a score. It's the same behavior as a lot of other functions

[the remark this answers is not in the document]
theogiraudet: So the result in the storage is not scaled?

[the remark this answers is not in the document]
aksiome: Yeah this way the function behave as a data get. If the user needs an int he can use a scale but the real value is stored in a double

docs/modules/environment.md
```diff
@@ -140,5 +360,132 @@ Determine if it is currently thundering or not.
...
+:::::
+
+---
+
+### Villager Behavior
+
+:::::{tab-set}
+::::{tab-item} Working
+
+**`bs.environment:are_villagers_working`**
```
aksiome, line 411: This name imply a state which is not really what is checked. Villagers could be awake in a sleeping phase if there is no bed for example. I suggest:
is_villager_sleep_time
is_villager_work_time
is_villager_meet_time
Or
*_sleeping_time
_working_time
_socializing_time

docs/modules/environment.md
```diff
@@ -140,5 +360,132 @@ Determine if it is currently thundering or not.
...
+
+Determine if undead mobs (zombies, skeletons, etc.) can currently burn in sunlight (between 05:27 and 18:32 in clear weather).
+
+> **Credits**: theogiraudet
+
+---
+
+### Is Monsters Spawn Period?
+
+**`bs.environment:is_monsters_spawn_period`**
```
aksiome, line 462: Why period instead of time if you use time everywhere else

docs/modules/environment.md
```diff
@@ -140,5 +360,132 @@ Determine if it is currently thundering or not.
...
+
+Determine if beds can currently be used by players (between 18:32 and 05:27 in clear weather, or 18:00 and 05:59 in rainy weather).
+
+> **Credits**: theogiraudet
+
+---
+
+### Can Undeads Burn?
+
+**`bs.environment:can_undeads_burn`**
```
aksiome, line 452: Same?
Is_undead_burning_time of is_undead_burn_time

docs/modules/environment.md
```diff
@@ -140,5 +360,132 @@ Determine if it is currently thundering or not.
...
+
+Determine if hostile monsters can currently spawn outdoors (between 19:11 and 04:48 in clear weather, or 18:58 and 05:01 in rainy weather).
+
+> **Credits**: theogiraudet
+
+---
+
+### Are Bees Sleeping?
+
+**`bs.environment:are_bees_sleeping`**
```
aksiome, line 472: Same

docs/modules/environment.md
```diff
@@ -140,5 +360,132 @@ Determine if it is currently thundering or not.
...
+
+Determine if bees are currently sleeping in their nests/hives (between 18:32 and 05:27 in clear weather).
+
+> **Credits**: theogiraudet
+
+---
+
+### Is Creakings Spawn Period?
+
+**`bs.environment:is_creakings_spawn_period`**
```
aksiome, line 482: Same as for monsters

modules/bs.environment/data/bs.environment/predicate/are_bees_sleeping.json
aksiome, line 1: I didn't initially anticipate that the predicate checks more than just the time (also weather), but I still think using the _time suffix is appropriate. Time can be interpreted in a broader sense to mean the overall context in which the behavior is expected to happen, not just the literal time of day.
One thing is sure are_bees_sleeping is not appropriate imo since the predicate doesnt really check the state of bees

docs/modules/environment.md
```diff
@@ -84,6 +84,226 @@ function #bs.environment:get_temperature {scale:1000}
...
+```{function} #bs.environment:get_sun_angle
+
+Get the sun's angle on the Y axis relative to the horizon at a specific time.
+
+:Inputs:
+  **Function macro**:
+  :::{treeview}
+  - {nbt}`compound` Arguments
+    - {nbt}`int` **day**: The day number.
+    - {nbt}`int` **daytime**: The time of day in ticks (0-24000).
```
aksiome, line 171: Im really not convinced by those params. It makes the function hard to use. Since those are ints i think we should use input scores for both daytime and time. Sure im assuming that scale is mostly static and that daytime and time are computed and change a lot but i think this is a fair assumption. Also even if we mix input types i think it remains a simple signature and avoid passing around a macro. This is a bit like in get and set pos in the position module where we use scores with the scale param. Last argument is that with the current impl it will encourage datapackers to use the internal function without calling the function tag (which is something i've already seen people doing and that i really think we should avoid)

aksiome (changes requested):

theogiraudet:
@aksiome better?

PR #480, opened 2025-07-20
✨  feat(bs.fsm): add bs.fsm module

theogiraudet:


aksiome:
To summarize what we discussed earlier:

The terms **`new`**, **`start`**, and **`instance`** are closely tied to object-oriented programming (OOP), which we should try to avoid in this context.
Personally, I prefer terminology like **`template`**, along with functions such as **`define`** and **`init`**.

---

### Updated Personal Proposal

I think we should primarily rely on **storage** instead. There’s no real need for a macro, directly using storage for the template is simpler for users and allows them to build or modify the FSM dynamically (across multiple commands or loops).
This approach also removes the need for a dedicated template name, leaving it to the user to handle namespacing.

We can drop the separate concepts of *new* and *start*, and instead use a single initialization function (e.g., `create`, `init`, or `start`), such as:

```mcfunction
function #bs.fsm:create {
    name: str,
    uses: str,
    with: {
        validate: true
    }
}
```

* **`uses`**, storage path to the template. (Possibly could also accept a compound or other data sources like entities?)
* **`validate`**, determines whether the FSM should be validated on creation. (personally not fan of it see last note)

We would also introduce a new function, **`bs.fsm:validate`**, allowing users to pre-validate a template instead of validating it with each new instance.

In my opinion (though I expect you might disagree), the user should be responsible for validating the template.
We could even remove the `validate` parameter entirely inside the create function. If we include validation in every example, it should be clear enough that it’s an essential step.

Also we could flag validated template and emit a warning if the user is using an un-validated template. What i'm not a fan of is that if the user creates a static template and validated it at design time there is actually no need to either validate it in load or when creating the fsm.

### Final State

We haven’t made a final decision on this yet, but I think we have two options:
either introduce the **`type`** field as you suggested, or, perhaps better if possible, add this parameter to the **validation function** instead. That approach would keep the FSM template minimal and ensure that all validation-related options are handled within the `validate` function.

PR #510, opened 2025-12-27
✨ feat(bs.collection): add bs.collection module

theogiraudet:
### Description
The goal of this module is to add numerous features about manipulating Minecraft arrays.
This module is NOT about providing ways to stream arrays.
The operators are not applied lazily.

All planned features:

# Collection Module Features

This document lists all features for the Bookshelf Collection module. Checked boxes indicate implemented features.

### Transformation Operations
- [x] **map** - Transform each element using a function
- [x] **flatmap** - Transform each element and flatten the results
- [x] **flatten** - Flatten a nested collection by one level

### Filtering Operations
- [x] **filter** - Keep only elements matching a predicate

### Validation Operations
- [x] **any** - Check if any element matches a predicate
- [x] **all** - Check if all elements match a predicate
- [x] **none** - Check if no elements match a predicate
- [x] **find** - Find first element matching a predicate
- [x] **find_last** - Find last element matching a predicate

### Aggregation Operations
- [x] **count** - Count the number of elements in a collection
- [x] **sum** - Calculate the sum of all numbers (uses scale 1000 for precision)
- [x] **min** - Find the minimum value in a collection
- [x] **max** - Find the maximum value in a collection
- [x] **average** - Calculate the average (mean) of numbers (uses scale 1000 for precision)
- [x] **reduce** - Reduce collection to a single value using an accumulator function
- [x] **reduce_right** - Reduce collection to a single value using an accumulator function, start from the right of the collection
- [x] **fold** - Reduce with a provided initial value
- [x] **fold_right** - Reduce with a provided initial value, start from the right of the collection

### Iteration Operations
- [x] **foreach** - Execute an operation on each element

### Utility Operations
- [x] **reverse** - Reverse the order of elements in a collection
- [x] **is_empty** - Check if collection is empty
- [x] **contains** - Check if collection contains element
- [x] **index_of** - Find index of first matching element
- [x] **last_index_of** - Find index of last matching element
- [x] **scale** - scale all numbers of a array of numbers

### Intermediate Operations
- [x] **distinct** - Remove duplicate elements
- [ ] **sorted** - Sort elements (with optional comparator)
- [x] **take** - Take first N elements
- [x] **drop** - Skip first N elements
- [x] **take_while** - Take elements while predicate is true
- [x] **drop_while** - Drop elements while predicate is true
- [x] **slice** - Extract slice from start to end index
- [x] **chunk** - Split into chunks of size N
- [x] **partition** - Split into two collections based on predicate
- [ ] **group_by** - Group elements by key function
- [x] **zip** - Combine two collections element-wise
- [x] **peek** - Execute side effect without consuming (returns collection)
- [x] **scanr** - Like reduce but returns intermediate results
- [x] **scanr_right** - Like reduce_right but returns intermediate results
- [x] **scanf** - Like fold but returns intermediate results
- [x] **scanf_right** - Like fold_right but returns intermediate results
- [x] **tap** - Execute callback on entire collection and return it

### Combination Operations
- [x] **concat** - Combine multiple collections sequentially
- [ ] **merge** - Merge collections with custom merge function
- [x] **intersect** - Get common elements between collections
- [x] **union** - Combine unique elements from multiple collections
- [x] **difference** - Elements in first but not in second collection
- [x] **symmetric_difference** - Elements in either but not both collections

### Advanced Operations
- [x] **sliding** - Sliding window over elements
- [ ] **associate** - Create compound from collection using key/value functions
- [ ] **associate_by** - Create compound using key function (value is element)
- [x] **flatten_deep** - Recursively flatten nested collections

### Generator
- [x] **range** - Generate a range of numbers
- [x] **repeat** - Generate a collection of repeated values
- [x] **generate** - Generate a collection of values using a generator function and a limit
- [x] **generate_while** - Generate a collection of values using a generator function and a predicate
- [ ] **from_compound** - Generate a collection from a compound


docs/modules/collection.md
aksiome: Im not sure if it's really a good thing to mention: This module is NOT a stream module, features are not lazy. Like yeah the name is collection now so of course it's not a stream. For me there is no ambiguity especially in MC. Imo the only ambiguity comes from the fact that we thought about making a stream module but people dont know that

aksiome (changes requested):
So far, I’ve mainly reviewed the tests and documentation, not the implementation itself. While looking at usage, I was thinking it could be useful to provide some pre-made callback functions for common predicate tests.
This way, instead of working directly with the storage, users could simply specify a range, a value, or other standard checks. I'm not considering this for this PR now, but it could be a nice addition later?

docs/modules/collection.md
theogiraudet: I don't think so. In the context of programming, higher-order functions such as map, filter, zip, etc. are most of the time used in the context of streams/lazy collections.

PR #530, opened 2026-02-27
🐛 fix(bs.environment): fix bs.environment for release 26.1

theogiraudet:


aksiome (approved):
LGTM! I think we should discuss a bit more about world clock in the future since they are dimensions dependant? Like does it work in the nether, and if not then maybe we could use function instead with macro and inline predicate?

PR #561, opened 2026-07-10
✨ feat(bs.geometry): add bs.geometry module

RacoonJohn:
### Description
Add a way to compute the inplane coord of a line/plane intersection


modules/bs.screen_detection/data/bs.screen_detection/function/system/load.mcfunction
```diff
@@ -0,0 +1,5 @@
+scoreboard objectives add bs.screen_detection dummy
+scoreboard players set #1000 bs.screen_detection 1000
+scoreboard players set #256 bs.screen_detection 256
+scoreboard players set #2 bs.screen_detection 2
```
theogiraudet: In Bookshelf, we are using `bs.const` for constant values.
Whe should reuse it here.

modules/bs.screen_detection/data/bs.screen_detection/function/api/get_plane.mcfunction
```diff
@@ -0,0 +1,6 @@
+function bs.screen_detection:system/safety/load
+
+execute summon marker run function bs.screen_detection:system/get_plane
+
+function bs.screen_detection:system/safety/unload
```
theogiraudet: I am not sure to correctly understand why your are loading/unloading the chunk at 0 0 while your entity seems to be summoned at the execution position (so most of the time, outside the 0 0 chunk).

If you need an entity at a fixed position, in a loaded chunk, so you can use the entity we provide: `B5-0-0-0-1`, which is also a marker (you should summon it if not present in your load.json).

modules/bs.screen_detection/data/bs.screen_detection/function/api/plane_coord.mcfunction
```diff
@@ -0,0 +1,23 @@
...
+scoreboard players operation #dot0 bs.screen_detection *= #u0 bs.screen_detection
+
+scoreboard players operation #dot1 bs.screen_detection = #n1 bs.screen_detection
+scoreboard players operation #dot1 bs.screen_detection *= #u1 bs.screen_detection
+scoreboard players operation #dot0 bs.screen_detection += #dot1 bs.screen_detection
+
+scoreboard players operation #dot1 bs.screen_detection = #n2 bs.screen_detection
+scoreboard players operation #dot1 bs.screen_detection *= #u2 bs.screen_detection
+scoreboard players operation #dot0 bs.screen_detection += #dot1 bs.screen_detection
+scoreboard players operation #dot0 bs.screen_detection /= #1000 bs.screen_detection
```
theogiraudet: For simple arithmetic operations, you can use `bs.ctx` instead, to avoid creating many objectives.

modules/bs.screen_detection/data/minecraft/tags/function/load.json
theogiraudet, line 1: No need to include this file, our build tool will generate it anyway.

modules/bs.screen_detection/README.txt
theogiraudet, line 1: We are using a Sphinx documentation, don't hesitate to create a new page for the new module, with the documentation of your feature inside.
To see how to contribute to the documentation: https://docs.mcbookshelf.dev/en/latest/contribute/documentation/

theogiraudet (commented):
Hi!

First, thank you very much for your contribution :)
I didn't read the semantics of the code yet, but here some preliminary suggestions:
- I believe this new feature is too generic for a UI-oriented module, I propose to put it in a `bs.geometry`, which can then be used by a future `bs.components` module for instance, just like `bs.interaction`.
- in Bookshelf, we don't use an `api` folder for the public API.
In fact, users should, most of the time, not use our mcfunction file at all.
Instead, we are using a `function tag` interface, so each feature that should be usable should declare a function tag: https://docs.mcbookshelf.dev/en/latest/contribute/metadata/#feature-metadata.
This is also the file citing you as the author of the feature ;)

In general, don' hesitate to check https://docs.mcbookshelf.dev/en/latest/contribute/ for our contribution, each contribution should respect it as most as possible :)
https://docs.mcbookshelf.dev/en/latest/contribute/

modules/bs.screen_detection/data/bs.screen_detection/function/api/get_line.mcfunction
theogiraudet, line 1: Each file must be prefixed by a copyright header.
This is the way for us to be sure that you accept the distribution of your code in our terms!
https://docs.mcbookshelf.dev/en/latest/contribute/file-format/

RacoonJohn:
Thank you for your help! I'll make the change today :)

theogiraudet:
I think we should also discuss about the interface of such a feature :)
After discussing with @VForiel, we think a two-parts module can be interesting:
- Geometry object creation:
  - `bs.geometry create_plane`
  - `bs.geometry create_line`
- Operations:
  - `bs.geometry intersection`
 
So close to what you are proposing.

The idea, is that `get_intersection_point` is able to identify the "type" of given objects, and delegate the computation to an internal function.
Like that, we could in the future add new types (`tube`, `cone`, `square`, …) and new operations (`union`, `symetric_difference`, `difference`, …).
Remain to define how the objects are given to the feature: naively I believe doing just like `bs.collections` and `bs.blocks`, and giving all through storage can be a good idea.
Finally, we should probably define the format of object themselves, here an idea:

```
{ type: "line", x1: 0, y1: 0, x2: 5, y2: 5 }
{ type: "plane": x1: 0, y1: 0, x2: 1, y2: 1, x3: 0, y3: 1 }
{ type: "point": x1: 0, y1: 0 } // return value for get_intersection_point when involving a line
```
But this will probably require a bit of discussion with everyone (cc @aksiome).

RacoonJohn:
Need to do more search but i guess sdf can be a way to achieve that + more search to compute the coord of a point in other spaces than a cartesian

but sound like a fun project to do!

theogiraudet:
Yeah, I think we can have funny things that can be useful, among other, to realize screens ;)

modules/bs.geometry/data/bs.geometry/function/get_line.mcfunction
```diff
@@ -0,0 +1,33 @@
...
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+
+
+
```
theogiraudet: Keep only one line between the header and the start of the code (too apply to other files too).

modules/bs.screen_detection/README.txt
theogiraudet, line 1: Since you have now the documentation, this file becomes useless, so can be deleted! :)
However we require a `README.md` which is used to define the displayed text on Modrinth and Smithed.
Look at other module to see the format!

modules/bs.screen_detection/data/bs.screen_detection/function/get_line.mcfunction
```diff
@@ -0,0 +1,33 @@
...
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+
+
+
+tp B5-0-0-0-1 ~ ~ ~
```
theogiraudet: Okay, if you need to move the entity, so you should probably use your own instead of the one we propose :)
Sorry for that, I perhaps misunderstood the previous version of the code.
However, you can follow the same format as ours for the entity's tags.

modules/bs.screen_detection/module.json
```diff
@@ -0,0 +1,16 @@
...
+  "extend": "../config.json",
+  "data_pack": {
+    "name": "bs.screen_detection",
+    "load": "."
+  },
+  "meta": {
+    "name": "Screen Detection",
+    "slug": "bookshelf-screen_detection",
+    "description": "compute the inplane coord of a line/plane intersection",
+    "documentation": "",
```
theogiraudet: You can add the link to the documentation page (cf other module.json to see the URL).

modules/bs.screen_detection/module.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extend": "../config.json",
+  "data_pack": {
+    "name": "bs.screen_detection",
```
theogiraudet: So do we agree that we rename this module by `bs.geometry`?

modules/bs.screen_detection/data/bs.screen_detection/function/__load__.mcfunction
```diff
@@ -0,0 +1,27 @@
...
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+
+forceload add -1 -1 0 0
+forceload add -30000000 1600
+execute unless entity B5-0-0-0-1 run summon minecraft:marker -30000000 0 1600 {UUID:[I;181,0,0,1],Tags:["bs.entity","bs.persistent","smithed.entity","smithed.strict"]}
+
+scoreboard objectives add bs.const dummy [{text:"BS ",color:"dark_gray"},{text:"Constants",color:"aqua"}]
+    scoreboard players set #1000 bs.const 1000
```
theogiraudet: Beware of this indentation :)

theogiraudet: For bs.const, we don't use the `#` before the name.
So just `1000` instead of `#1000`.

modules/bs.geometry/data/bs.geometry/function/__load__.mcfunction
```diff
@@ -0,0 +1,27 @@
...
+# Conditions:
+# - You may use this file in compliance with the MPL v2.0
+# - Any modifications must be documented and disclosed under the same license
+#
+# For more details, refer to the MPL v2.0.
+# ------------------------------------------------------------------------------------------------------------
+
+
+forceload add -1 -1 0 0
+forceload add -30000000 1600
```
theogiraudet, line 17: You probably need an `__unload__` to remove scoreboards and forceload.

theogiraudet:
Since this is your first contribution, we will probably iterate a bit to have something meeting our conventions.
I hope that's not a problem for you :)

modules/bs.screen_detection/data/bs.screen_detection/function/__load__.mcfunction
RacoonJohn: i'm used to add an identation when i initialize fakeplayer scores in load function, is it an issue?

theogiraudet: Hum not really I suppose.
This is mainly to uniform the code with the rest of the codebase.

modules/bs.screen_detection/module.json
RacoonJohn: if other contributors agree to it i don't have any issue with that

theogiraudet: @VForiel @aksiome

aksiome:
Quick thing about score naming. All score holders except for `bs.ctx` should be prefixed by the module name (without `bs`) and the feature, so for example `$geometry.get_plane.dx bs.out`. The other thing is that `bs.ctx` should use only a single letter, so `#a, #b, ... #x, #y, #z`.

Honestly, that last one is a bit painful to write code with, but there's a rationale. The first rule keeps a clear separation between all modules and features for the user. The second (`bs.ctx`) is about reusing scores internally as much as possible and avoiding too many entries in a single objective, at some point we noticed that had an impact on performance.

aksiome (changes requested):
EDIT: I took the liberty of updating the branch myself, so you just need to fetch it locally.


Thanks for the module ! I finally had some time to do a proper review, it's in great shape, and I really like what you did with the API.

## Blocking

### 1. Malformed function headers

Both of these need their headers fixed:

- `modules/bs.geometry/data/bs.geometry/function/transform/coord_space/spheric_to_cartesian.mcfunction`
- `modules/bs.geometry/data/bs.geometry/function/transform/coord_space/cylindric_to_cartesian.mcfunction`

### 2. Missing tests

We've switched to [ward](https://github.com/mcbookshelf/ward) for testing, so please update your branch to master, you'll also get the correct Spyglass autocompletion from it. The command reference is here: https://github.com/mcbookshelf/ward/blob/main/docs/commands.md

We don't need exhaustive coverage, just a few tests so the module is verified and a future change can't break it silently. To run them locally:

```sh
uv run modules test bs.geometry
```

### 3. Documentation not registered

The docs need to be added to the runtime toctree section of `modules/index.md`, and the module should be renamed to **Geometry**. You can check the result with:

```sh
uv run docs watch
```

It should then show up in the left sidebar on the modules page.

### 4. Function tags should be flat

We don't nest function tags in folders except for functions that aren't features (callbacks, for example). Internal logic can stay structured however you like, only the user-facing tags matter here:

- `#bs.geometry:get_point`
- `#bs.geometry:get_line`
- `#bs.geometry:get_plane`
- `#bs.geometry:get_sphere`
- `#bs.geometry:get_cartesian_space`
- `#bs.geometry:get_cylindric_space`
- `#bs.geometry:get_spherical_space`
- `#bs.geometry:intersect`
- `#bs.geometry:project_ortho`
- `#bs.geometry:rotate_axis`
- `#bs.geometry:convert_space`
- `#bs.geometry:get_signed_distance`

These names are suggestions, if you think some of them aren't ideal, I'm open to better ones.

### 5. Error handling

I'm not fond of the error object. What we usually do in this case is log the error through the log module (see `sidebar/create` for an example). Returning fail alongside it would be a good addition too.
That means also updating the `module.json` to include `bs.log` as a weak dependency.

### 6. `get_sphere` interface

`get_sphere` is the only function using a score. We try not to mix interface types (storage, macro, score) within a single module. Storage and macro also remove the need for a scaling factor and let the user pass a plain float/double, which means changing the internal scaling later won't break anything from the outside.

My preference here would be a macro, since the radius is mostly static and unlikely to change often, so cache misses should be low:

```mcfunction
function #bs.geometry:get_sphere {radius:1.78}
```

That said, I'll let you pick whichever you prefer.

### 7. Undocumented particle function

There's a particle function that isn't documented, if it's only there for testing, let's remove it from the module.

### 8. snake_case in storage paths

I noticed `bs:out geometry.Pos`. We use snake_case everywhere, even where Minecraft doesn't (they're moving that way themselves anyway), so this should be `bs:out geometry.pos`. I may have missed other places, but you basically should never use an uppercase (except in comments of course)

### 9. Entity handling

I'm unsure why custom entities are used here when Bookshelf provides its own. It's better to reuse `B5-0-0-0-1` for example to avoid spawning too many entities. There's also a problem with the current marker setup: teleporting an entity across dimensions will fail. I'd strongly suggest this pattern instead:

```mcfunction
execute summon marker run function get_pos
```

with `kill @s` inside the function. It's slightly less performant, but it's the only approach that works correctly in every dimension.

## Non-blocking

These either don't need to hold up the PR or need a bit more discussion.

- **`pack.png`**, one is required, but don't worry about it, I can contribute it myself.
- **Doc formatting**, the format doesn't fully match our other docs, mainly the use of treeview to display NBT blocks. Happy to update the docs myself if you'd rather not deal with it.
- **Shape format section**, a short section describing the shape format would be nice, so people can build shapes directly in storage without going through the helper functions that create them from the world.


That's it for me, but maybe @theogiraudet will have additional things to say !

RacoonJohn:
I'll correct the issues as soon as i can! For Entity handling, the teleportation across dimension is not usefull, i use the coordonates  context to put it as the origin vector of the returned shape, there is no entity or block placed at the position. So i'll just swotch it to B5-0-0-0-1

For Non Blocking, I'll like a lot your help with `pack.png` and doc formatting (it's the first time i properly participate to a collaborative project and i have a bit of difficulties but it's i'm learning a lot of usefull skills !) 

I'll do my best to shape format doc but there is somes maybe weird things i made, f.e the sphere get from get_sphere have `i`, `j`, `k` director vectors to work with `orth_projection` , `rot_axis` and `coord_space` transformations as if it is an 
orthonormal coordinate system, but these director  vector are not used for `interesection` or `sdf`, if you think that it should only have necessary fields in shape format it'll be surely more understandable (and a bit more efficient for get_sphere and other 3D non directed shapes) but imo will add some struggle to use theses transformations and less uniformisation accross different shapes format

aksiome:
Im thinking for the pack.png maybe something like a triangle ruler (looks also like an angle)
bookshelf-book-geometry

theogiraudet:
Personally, I like it.

RacoonJohn:
Me too

aksiome:
> For Entity handling, the teleportation across dimension is not usefull, i use the coordonates context to put it as the origin vector of the returned shape, there is no entity or block placed at the position. So i'll just swotch it to B5-0-0-0-1

If it's only for rotation it's fine, if it's only for let's say a unit vector we just need to add `in minecraft:overworld` to force the commands to run in the overworld (you cannot assume all people will always only use the module in the overworld and not in the nether for example). If it's for coordinate grabbing, then there is a good chance we are stuck with a marker instead, because `in overworld` will multiply coordinates if in the nether.

> I'll do my best to shape format doc but there is somes maybe weird things i made, f.e the sphere get from get_sphere have `i`, `j`, `k` director vectors to work with `orth_projection` , `rot_axis` and `coord_space` transformations as if it is an
> orthonormal coordinate system, but these director vector are not used for `interesection` or `sdf`, if you think that it should only have necessary fields in shape format it'll be surely more understandable (and a bit more efficient for get_sphere and other 3D non directed shapes) but imo will add some struggle to use theses transformations and less uniformisation accross different shapes format

What im thinking here is that each function should only have what is needed for the function to run, but we should have a full section dedicated on how to create those shapes directly with the storage (without `get_` functions), and here we should specify the full format for valid shapes. Also if you need a bit of help to format things inside the doc we can rework it a bit later, but you can also take inspiration from other modules that use the storage and directly see how it looks with `uv run docs watch`.

RacoonJohn:
The use of `execute summon` now make sens for me, will commit the change soon!
I'll also change somes `get_` functions to have only what is needed for it
For the docs i'll try to do as much as i can to format it with the inspiration of the other modules but i'll need a bit of help
(I also planned to add more comments to the code itslef)

Issue #574, opened 2026-09-05
Outdated bs.health


RacoonJohn:
### 🗂️ Task Type

📖 Example

### 📝 Description

bs.health module is outdated with 26.3 upcoming, optimizations can be mades

PR #577, opened 2026-09-05
⚡ perf(bs.health): optimize health computation

RacoonJohn:
### Description
Use compute and float number provider to do operations 
Still use scoreboards to store players data (scale of 100000)

### Related Issues
[bs.health outdated](https://github.com/mcbookshelf/bookshelf/issues/574)

### Additional Notes
I think somes macros could be change to storage inputs and save performances

### DIFF
--- a/modules/bs.health/data/bs.health/function/add/health.mcfunction
+++ b/modules/bs.health/data/bs.health/function/add/health.mcfunction
@@ -15,19 +15,14 @@
 
 # Note: Thanks to XanBelOr for the idea of using the effects_changed trigger advancement
 
-# Get current health, max_health, and input points
-execute store result score #h bs.ctx run data get entity @s Health 1000000
-scoreboard players add #h bs.ctx 5
-scoreboard players operation #h bs.ctx /= 10 bs.const
-execute store result score #m bs.ctx run attribute @s minecraft:max_health get 100000
-$execute store result score #p bs.ctx run data get storage bs:const health.point $(points)
+# Get current health, max_health
+data modify storage bs:ctx h set from entity @s Health
+data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
 
 # Add incoming points to the healing modifier and clamp to max possible healing
-scoreboard players operation @s bs.hmod += #p bs.ctx
-scoreboard players operation #m bs.ctx -= #h bs.ctx
-scoreboard players operation @s bs.hmod < #m bs.ctx
-scoreboard players operation #m bs.ctx += #h bs.ctx
+$data modify storage bs:ctx f set compute default float {type:"min",inputs:[{type:"add",inputs:[$(points)f,{type:"mul",inputs:[0.00001,{type:"from_int",input:{type:"score",target:"this",score:"bs.hmod"}}]}]},{type:"add",inputs:[{type:"storage",storage:"bs:ctx",path:"m"},{type:"negate",input:{type:"storage",storage:"bs:ctx",path:"h"}}]}]}
+execute store result score @s bs.hmod run data get storage bs:ctx f 100000
 
 # Apply health change: reduction is instant, increase waits for instant_health to take effect
 execute if score @s bs.hmod matches ..-1 run return run function bs.health:utils/decrease_health
-execute if score @s bs.hmod matches 1.. run return run function bs.health:utils/increase_health
+execute if score @s bs.hmod matches 0.. run return run function bs.health:utils/increase_health
--- a/modules/bs.health/data/bs.health/function/add/health.mcfunction
+++ b/modules/bs.health/data/bs.health/function/add/health.mcfunction
@@ -16,12 +16,12 @@
 # Note: Thanks to XanBelOr for the idea of using the effects_changed trigger advancement
 
 # Get current health, max_health
-data modify storage bs:ctx h set from entity @s Health
-data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
+data modify storage bs:ctx x set from entity @s Health
+execute store result storage bs:ctx z float 0.00001 run attribute @s max_health base get 100000
 
 # Add incoming points to the healing modifier and clamp to max possible healing
-$data modify storage bs:ctx f set compute default float {type:"min",inputs:[{type:"add",inputs:[$(points)f,{type:"mul",inputs:[0.00001,{type:"from_int",input:{type:"score",target:"this",score:"bs.hmod"}}]}]},{type:"add",inputs:[{type:"storage",storage:"bs:ctx",path:"m"},{type:"negate",input:{type:"storage",storage:"bs:ctx",path:"h"}}]}]}
-execute store result score @s bs.hmod run data get storage bs:ctx f 100000
+$data modify storage bs:ctx w set compute default float {type:"min",inputs:[{type:"add",inputs:[$(points)f,{type:"mul",inputs:[0.00001,{type:"from_int",input:{type:"score",target:"this",score:"bs.hmod"}}]}]},{type:"add",inputs:[{type:"storage",storage:"bs:ctx",path:"m"},{type:"negate",input:{type:"storage",storage:"bs:ctx",path:"h"}}]}]}
+execute store result score @s bs.hmod run data get storage bs:ctx w 100000
 
 # Apply health change: reduction is instant, increase waits for instant_health to take effect
 execute if score @s bs.hmod matches ..-1 run return run function bs.health:utils/decrease_health
--- a/modules/bs.health/data/bs.health/function/add/health.mcfunction
+++ b/modules/bs.health/data/bs.health/function/add/health.mcfunction
@@ -25,4 +25,4 @@ execute store result score @s bs.hmod run data get storage bs:ctx f 100000
 
 # Apply health change: reduction is instant, increase waits for instant_health to take effect
 execute if score @s bs.hmod matches ..-1 run return run function bs.health:utils/decrease_health
-execute if score @s bs.hmod matches 0.. run return run function bs.health:utils/increase_health
+execute if score @s bs.hmod matches 1.. run return run function bs.health:utils/increase_health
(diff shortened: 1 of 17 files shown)


modules/bs.health/data/bs.health/function/add/health.mcfunction
```diff
@@ -15,19 +15,14 @@
...
 
-# Get current health, max_health, and input points
-execute store result score #h bs.ctx run data get entity @s Health 1000000
-scoreboard players add #h bs.ctx 5
-scoreboard players operation #h bs.ctx /= 10 bs.const
-execute store result score #m bs.ctx run attribute @s minecraft:max_health get 100000
-$execute store result score #p bs.ctx run data get storage bs:const health.point $(points)
+# Get current health, max_health
+data modify storage bs:ctx h set from entity @s Health
+data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
```
aksiome: I think we should keep the attribute command here. I don't think it will ever overflow, and this avoids having to deserialize the entity data twice?

Also, since we only read Health once in `bs:ctx h`, maybe we could reference the entity's health nbt directly in the compute command and avoid copying it to storage?

modules/bs.health/data/bs.health/function/add/health.mcfunction
```diff
@@ -15,19 +15,14 @@
...
-scoreboard players operation #m bs.ctx -= #h bs.ctx
-scoreboard players operation @s bs.hmod < #m bs.ctx
-scoreboard players operation #m bs.ctx += #h bs.ctx
+$data modify storage bs:ctx f set compute default float {type:"min",inputs:[{type:"add",inputs:[$(points)f,{type:"mul",inputs:[0.00001,{type:"from_int",input:{type:"score",target:"this",score:"bs.hmod"}}]}]},{type:"add",inputs:[{type:"storage",storage:"bs:ctx",path:"m"},{type:"negate",input:{type:"storage",storage:"bs:ctx",path:"h"}}]}]}
+execute store result score @s bs.hmod run data get storage bs:ctx f 100000
 
 # Apply health change: reduction is instant, increase waits for instant_health to take effect
 execute if score @s bs.hmod matches ..-1 run return run function bs.health:utils/decrease_health
-execute if score @s bs.hmod matches 1.. run return run function bs.health:utils/increase_health
+execute if score @s bs.hmod matches 0.. run return run function bs.health:utils/increase_health
```
aksiome: Is there a reason to use 0 that I might be missing? If the modified health is 0, why not just do nothing?

modules/bs.health/data/bs.health/function/add/max_health.mcfunction
```diff
@@ -14,9 +14,7 @@
 # ------------------------------------------------------------------------------------------------------------
 
 # Get input points and base max_health
-$execute store result score #p bs.ctx run data get storage bs:const health.point $(points)
-execute store result score #m bs.ctx run attribute @s minecraft:max_health base get 100000
+data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
```
aksiome: Same as before with the attribute command

modules/bs.health/data/bs.health/function/get/health.mcfunction
```diff
@@ -13,8 +13,5 @@
 # For more details, refer to the MPL v2.0.
 # ------------------------------------------------------------------------------------------------------------
 
-execute store result score #h bs.ctx run data get entity @s Health 1000000
-scoreboard players add #h bs.ctx 5
-scoreboard players operation #h bs.ctx /= 10 bs.const
-execute store result storage bs:out health.get_health double 0.00001 run scoreboard players operation #h bs.ctx += @s bs.hmod
-$return run data get storage bs:out health.get_health $(scale)
+data modify storage bs:ctx h set from entity @s Health
```
aksiome: Same as before, maybe reference it directly?

modules/bs.health/data/bs.health/function/get/max_health.mcfunction
```diff
@@ -13,5 +13,6 @@
 # For more details, refer to the MPL v2.0.
 # ------------------------------------------------------------------------------------------------------------
 
-execute store result storage bs:out health.get_max_health double 0.00001 run attribute @s minecraft:max_health base get 100000
-$return run data get storage bs:out health.get_max_health $(scale)
+data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
```
aksiome: Same as before with the attribute command

modules/bs.health/data/bs.health/function/set/health.mcfunction
```diff
@@ -15,16 +15,13 @@
...
 
-# Get current health, max_health, and input points
-execute store result score #h bs.ctx run data get entity @s Health 1000000
-scoreboard players add #h bs.ctx 5
-scoreboard players operation #h bs.ctx /= 10 bs.const
-execute store result score #m bs.ctx run attribute @s minecraft:max_health get 100000
-$execute store result score @s bs.hmod run data get storage bs:const health.point $(points)
+# Get current health, max_health
+data modify storage bs:ctx h set from entity @s Health
+data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
```
aksiome: Same as before with the attribute command and the Health ref

modules/bs.health/data/bs.health/function/ttl/normalize_unit.mcfunction
```diff
@@ -13,5 +13,5 @@
 # For more details, refer to the MPL v2.0.
 # ------------------------------------------------------------------------------------------------------------
 
-$execute store result storage bs:ctx _.time int $(scale) run data get storage bs:ctx _.time
+data modify storage bs:ctx _.time set compute default int {type:"mul",inputs:[$(scale),{type:"storage",storage:"bs:ctx",path:"_.time"}]}
```
aksiome: This should be a macro command since $(scale) is passed. I guess the initial idea was to save a macro in the calling function. We could just multiply `bs:const health.units[{name:'$(unit)'}]` and `bs:ctx _.time` in `register_unit`? (but we should keep the error log for when a unit is not supported)

modules/bs.health/data/bs.health/function/add/health.mcfunction
```diff
@@ -15,19 +15,14 @@
...
+# Get current health, max_health
+data modify storage bs:ctx h set from entity @s Health
+data modify storage bs:ctx m set from entity @s attributes[{id:"minecraft:max_health"}].base
 
 # Add incoming points to the healing modifier and clamp to max possible healing
-scoreboard players operation @s bs.hmod += #p bs.ctx
-scoreboard players operation #m bs.ctx -= #h bs.ctx
-scoreboard players operation @s bs.hmod < #m bs.ctx
-scoreboard players operation #m bs.ctx += #h bs.ctx
+$data modify storage bs:ctx f set compute default float {type:"min",inputs:[{type:"add",inputs:[$(points)f,{type:"mul",inputs:[0.00001,{type:"from_int",input:{type:"score",target:"this",score:"bs.hmod"}}]}]},{type:"add",inputs:[{type:"storage",storage:"bs:ctx",path:"m"},{type:"negate",input:{type:"storage",storage:"bs:ctx",path:"h"}}]}]}
```
aksiome: Also i would rather we stick to `bs:ctx x` `bs:ctx y` or `bs:ctx z` to keep the bs:ctx really small (and reuse the same keys as everywhere else)

modules/bs.health/data/bs.health/function/utils/decrease_health.mcfunction
```diff
@@ -13,13 +13,9 @@
 # For more details, refer to the MPL v2.0.
 # ------------------------------------------------------------------------------------------------------------
 
-scoreboard players operation #h bs.ctx += @s bs.hmod
-execute if score #h bs.ctx matches ..0 unless predicate {type:"entity_properties",entity:"this",predicate:{"type_specific/player":{gamemode:["creative","spectator"]}}} run kill @s
+data modify storage bs:ctx h set compute default float {type:"add",inputs:[{type:"storage",storage:"bs:ctx",path:"h"},{type:"storage",storage:"bs:ctx",path:"f"}]}
```
aksiome: I think there was a reason to keep `@s bs.hmod`. Let say the user fist add_health (which dont run immediatly) then just in the same tick remove more than was added. We need to substract that, which is why hmod is used instead of direct user input.

aksiome (changes requested):

modules/bs.health/data/bs.health/function/add/health.mcfunction
RacoonJohn: the score could be rounded to 0 but the data greater, like 0.000009

modules/bs.health/data/bs.health/function/ttl/normalize_unit.mcfunction
RacoonJohn: i think we should too, but i didn't know if this will break something

modules/bs.health/data/bs.health/function/utils/decrease_health.mcfunction
RacoonJohn: when computing f (equivalent of `hmod`) we get the score value in the compute to be exactly like `@s hmod` so there should not be any problem

modules/bs.health/data/bs.health/function/add/health.mcfunction
aksiome: i think it's best to only "trust" the score here for everything and have a small precision loss. A single source of truth. The heal gets applied after anyway through the advancement using the score, which mean the loss is here anyway.

modules/bs.health/data/bs.health/function/add/health.mcfunction
RacoonJohn: since the attribute command will cost a lost in the precision so i thought it could be better, but in other case i think the attribute command will be more performant

In context float provider i didn't find a way to get an entity nbt

aksiome: > In context float provider i didn't find a way to get an entity nbt

Maybe there isn't, i'm not completly up to date with everything possible now, also yeah with the precision loss i think the scaling is actually enough x) Like even with float we'd get some imprecision because well it's float32

Issue #578, opened 2026-09-05
bs.color outdated docs


RacoonJohn:
### 🗂️ Task Type

📚 Documentation

### 📝 Description

The doc of the color module is outdated with "ArmorItems" and item tag

Issue #576, opened 2026-09-05
bs.color don't support rgba


RacoonJohn:
### 🗂️ Request Type

Feature

### 🧩 Feature or Module Name

bs.color

### 📝 Description

All bs.color's functions don't support alpha in hex int and rgba format

Issue #575, opened 2026-09-05
bs.color outdated


RacoonJohn:
### 🗂️ Task Type

📖 Example

### 📝 Description

With the upcoming 26.3, bs.color can have somes improvements and optimizations

PR #579, opened 2026-09-05
♻️ refactor(bs.color)!: update to v5

RacoonJohn:
### Description
- Compute and float context provider improvement
- rgba support
- updated docs

### Related Issues
- [26.3 update](https://github.com/mcbookshelf/bookshelf/issues/575)
- [rgba support](https://github.com/mcbookshelf/bookshelf/issues/576)
- [outdated docs](https://github.com/mcbookshelf/bookshelf/issues/578)


aksiome:
A few things, since I noticed you’ve already updated this:

* `__main__` should never accept a macro. Base functions now only work with storage, but a macro function will be generated automatically. If the generated function can be made more efficient, you can also implement it directly in `__macro__`. However, if it’s only needed to pass data from the macro into storage, the generated `__macro__` already handles this.
* Changelogs are now split per module, as you can see in `health`. We still need to gather the full history from all the old changelogs, though, so if you don’t feel like doing that, I’m happy to take care of it myself.
* We now require a `module.bs` file. You can look at the other modules for the format, as well as the new vscode extension. This lets us define metadata and inputs in a single place, and also allows us to generate mcdoc on the fly. It enables some autodoc as well, so have a look at the `health` docs to see how the documentation page itself can be simplified a bit.
* Tests should avoid using "raw" functions. For the storage version, use `#bs.<module>:<feature>` directly, and for the macro version if there is one (some functions don't have inputs) use `#bs.<module>:<feature>.in`. Don't worry about the tags those are automatically generated.

aksiome:
Also i've noticed you haven't updated your branch to v5 so you don't have the latest things, just did it for you

PR #588, opened 2026-09-18
🐛 fix(bs.raycast): prevent entity ID collisions between casts

### LINKED ISSUE #586: bs.raycast can invoke entity callbacks on the wrong entity after consecutive casts
### 🔢 Version

v4.2.0

### 🔍 Related Feature

#bs.raycast:run

### ✅ Expected Behavior

Entity callbacks should always run as the entity that produced the recorded hit.

### ❌ Actual Behavior

`#bs.raycast:run` resets `#count bs.raycast.id` to `0` for every cast, but entities can retain positive `bs.raycast.id` values from previous casts.

`record/entity/add` only allocates a new ID if the entity does not already have a positive one. This means a retained ID can collide with an ID newly allocated during the next cast.

Later, `react/any` resolves recorded hits by selecting a `bs.raycast.checked` entity with the matching ID and `limit=1`, so an ID collision can make the callback run as the wrong entity.

There is also an early-return case with zero-sized custom hitboxes: the entity is tagged `bs.raycast.checked` before the hitbox check returns, so it can remain checked while retaining a stale ID.

### 🔁 Steps to Reproduce

1. In one cast, let entity A receive `bs.raycast.id = 1`.
2. Start another cast. `#count bs.raycast.id` is reset to `0`, but A keeps ID `1`.
3. A is checked again and keeps its existing positive ID, so the counter is not advanced.
4. A new entity B is hit and receives newly allocated ID `1`.
5. A and B are now both checked with the same ID.
6. When a hit with ID `1` is processed, the callback can resolve to the wrong entity.

The same issue can occur with a zero-sized custom hitbox that returns early after being tagged checked while retaining a stale ID.

### 💡 Possible Fix

Assign a fresh cast-local `bs.raycast.id` whenever an entity is first checked in `check/entity/any`, before any hitbox early return.

Since checked entities are already excluded from being checked again during the same cast, this happens once per candidate per cast.

The conditional ID allocation can then be removed from `record/entity/add`.

Bafy78:
### Description

Assigns a fresh cast-local `bs.raycast.id` when an entity is first checked, before any hitbox path can return early.

This prevents IDs retained from previous casts from colliding with newly allocated IDs and causing entity callbacks to resolve to the wrong entity.

The conditional ID allocation in `record/entity/add` is removed, and regression tests cover both consecutive casts with changing targets and zero-sized custom hitboxes.

### Related Issues

Closes #586

### Additional Notes

No public API changes are introduced, so no user-facing documentation changes are needed.

No changelog entry was added because v4.2.0 has already been released and there is currently no unreleased changelog on `master`.

Validated with:

- `uv run check`
- `uv run modules test bs.raycast`
- `uv run modules test`


aksiome (changes requested):
Everything is fine, we just need a changelog file (`docs/changelog/v4.2.1`).

Something like:
```
## 📦 [Bookshelf - 4.2.1](https://github.com/mcbookshelf/bookshelf/releases/tag/v4.2.1)

> **🩹 Hotfix!**
>
> *This is a hotfix for version 4.2.0.*


### `🔦 bs.raycast`

- 🐛 **[#586](https://github.com/mcbookshelf/bookshelf/issues/586)** - Prevent entity ID collisions between casts.
```

Also feel free to bump the version to 4.2.1 in `src/bookshelf/definitions.py` this will trigger nightly builds and we'll be able to do a proper release after.

aksiome (approved):

Issue #595, opened 2026-09-21
bs.raycast outdated


RacoonJohn:
### 🗂️ Task Type

Improvement

### 📝 Description

With the new compute fonctionalities, the raycast could be more precise

Issue #587, opened 2026-09-18
Modules remaining for the v5 rework


aksiome:
- [ ] animation
- [ ] bitwise
- [ ] block
- [ ] collection
- [ ] color
- [ ] dump
- [ ] environment
- [ ] fsm
- [ ] generation
- [ ] health
- [ ] hitbox
- [ ] id
- [ ] interaction
- [ ] link
- [ ] log
- [ ] math
- [ ] move
- [ ] position
- [ ] random
- [ ] raycast
- [ ] schedule
- [ ] sidebar
- [ ] spline
- [ ] string
- [ ] time
- [ ] tree
- [ ] vector
- [ ] view
- [x] xp

RacoonJohn:
- [ ] animation (aksiome)
- [x] bitwise (aksiome)
- [ ] block (aksiome)
- [ ] collection**
- [ ] color* (RacoonJohn)
- [ ] dump - slowly deprecate?
- [ ] environment - in need of a redesign with environment attributes
- [ ] fsm (theogiraudet?)
- [ ] generation (aksiome)
- [ ] geometry (RacoonJohn) - to rebase into v5, reopen and finish there
- [x] health (aksiome)
- [ ] hitbox (aksiome)
- [ ] id*
- [ ] interaction*
- [ ] link*
- [ ] log - slowly deprecate?
- [ ] math**
- [ ] move (aksiome) - need hitbox
- [ ] position*
- [ ] random (aksiome) - need generation for a good redesign and test of fractal and simplex noises
- [ ] raycast (aksiome) - need hitbox
- [ ] schedule*
- [ ] sidebar*
- [ ] spline (aksiome)
- [ ] string*
- [ ] time*
- [ ] tree*
- [ ] vector*
- [ ] view (aksiome) - need raycast / hitbox for some features
- [x] xp (aksiome)

* Small modules to prioritize (easy for anyone to pick up)
** Large modules that should be done across multiple PRs, but have simple features that anyone can contribute to

PR #596, opened 2026-09-21
update bs.raycast

RacoonJohn:
### Description
Use the 26.3 compute to have more accuracy and (maybe) performances from raycast

### Related Issues
[bs.raycast outdated](https://github.com/mcbookshelf/bookshelf/issues/595)

### Additional Notes
Some ward test don't work but it seems to work perfectly when i manually recreate the test in my MC world


aksiome:
Thanks for the PR!

Starting from now, we should target the `v5` branch rather than `master` for newer changes. The scripts are slowly starting to stabilize, so I think it’s becoming safe for people to contribute to updating modules there. For now, I think it would be better to focus on smaller modules while we’re still figuring things out.

There’s already a list of modules that need to be updated in #587. If possible, I’d like people to mention in advance which module they’re planning to work on. I’ve already started working on raycast myself, so it would be good to avoid accidentally duplicating work. Just leave a comment on the issue and I’ll add your name next to the module you’re working on. I’ll definitely take some inspiration from this PR, but I think I’d prefer to continue with my current work on it for now. There are a few other things I want to address at the same time, such as allowing a raycast to be run from inside a raycast callback (higher-order functions). Also it's one of our most important module so i want to do a ton of benchmarks on it and since it is tightly coupled with hitbox, it's gonna take a bit of time for me to be happy about the state of these two modules combined before i'll push anything.

As for the contribution guidelines, `v5` brings quite a few changes that I haven’t had time to properly document yet. I think the health module is currently a good example of how a module should be structured, and `uv run check` now validates more things as well.

About the failing test, it’s most likely because it’s running with an older version of Beet that doesn’t include number providers yet. For animation, for example, I had to add a small dataclass in the Beet plugin to load them properly. Beet was updated two days ago, though, so with the newer version it should work fine now.

